PR #23, opened 2021-02-05
Integrate @webb-tools/sdk-mixer with react hooks [no-ui]

shekohex:
this PR will enable us to use the Mixer as easily as calling `useMixer()` hook.


packages/react-environment/src/MixerProvider.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const { api, connected } = useApi();
+  const [mixerGroups, setMixerGroups] = useState<MixerAssetGroup[]>([]);
+
+  useEffect(() => {
+    if (!connected) return;
+    try {
+      api.query.mixer.mixerGroups
+        .entries()
+        .toPromise()
+        .then((result) => {
```
shekohex, line 33: @drewstone this call never resolves (on the anon local node)! 
any ideas?

packages/react-environment/src/MixerProvider.tsx
```diff
@@ -0,0 +1,69 @@
...
+        ),
+      });
+    }
+
+    setMixerGroups([]);
+  }, [api, connected]);
+
+  const init = useCallback(() => {
+    console.log(`Mixer initialized with ${mixerGroups}`);
+    return Mixer.init([new MixerAssetGroup(0, 'EDG', 32)]);
```
shekohex, line 56: added dummy MixerAssetGroup here for just testing, but this should be fetched from the `api`.

packages/react-environment/src/MixerProvider.tsx
drewstone: Have you tried using `useCall`? https://github.com/webb-tools/webb-dapp/blob/master/packages/react-hooks/src/useCall.ts
```
  const mixerGroups = useCall<MixerGroup[]>('query.mixer.mixerGroups, []);
```

PR #32, opened 2021-02-16
Page Features and Mixer inti

AhmedKorim:
This PR covers those changes

- Prevent showing the page if the chain is not supported, Thus underlying components shouldn't be called if the features aren't met.
- Initializing the mixer, and storing, consuming an old mixer will be on another PR


drewstone:
Looks good but I get this when I run it!
image

drewstone (changes requested):
While I do see the error message, the app still crashes afterwards on my end.

PR #37, opened 2021-02-17
Deposit Page UI implementation

### LINKED ISSUE #7: Implement deposit page
Screen Shot 2021-01-28 at 11 22 14

Instead of an input for deposit amount, it should be a set of radio buttons that each fetch the size of the mixer similar to 

image

AhmedKorim:
Closes #7


packages/page-mixer/src/components/deposit/DepositConsole.tsx
```diff
@@ -1,107 +1,132 @@
...
+  const { allCurrencies } = useConstants();
 
+  const handleTokenCurrencyChange = useCallback(
+    (currency: CurrencyId) => {
+      setToken({ token: currency });
+      clearAmount();
+    },
+    [setToken, clearAmount]
+  );
+  const items = useMemo(
```
drewstone, line 78: We should pull these values from the mixerGroups directly. It should be available on each group from `mixerInfo. fixed_deposit_size`

packages/page-mixer/src/components/deposit/index.tsx
```diff
@@ -1 +1 @@
-export * from './DepositConsole';
+export * from './WithdrawConsole';
```
drewstone, line 1: This seems improperly named since it's in the deposit folder?

packages/page-mixer/src/components/withdraw/index.tsx
```diff
@@ -1 +1 @@
-export * from './WithdrawConsole';
+export * from './DepositConsole';
```
drewstone, line 1: Again this seems reversed?

packages/page-mixer/src/index.tsx
```diff
@@ -49,10 +63,13 @@ const PageSwap: FC = () => {
...
+        <WithdrawConsole />
       </Col>
     </Row>
   );
 };
 
-export default PageSwap;
+export default pageWithFeatures({
+  features: ['mixer'],
+  message: "Mixer isn't supported on the current cain ,please consider change the current network",
```
drewstone, line 74: "The mixer module is not supported on the current chain, please change the current network."

packages/react-components/src/utils/FeaturesGuard/pageWithFeatures.tsx
```diff
@@ -0,0 +1,34 @@
...
+    return <Component {...props} />;
+  }
+
+  return (
+    <Row gutter={[0, 24]}>
+      <Col span={24}>
+        <Information
+          variant={'warning'}
+          content={
+            message || "This feature isn't supported on the current chain, please consider change the current network"
```
drewstone, line 27: "please consider changing the current network"

packages/react-hooks/src/mixer/useMixer.ts
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  const [called, setCalled] = useState(false);
+
+  const init = useCallback(async () => {
+    setCalled(true);
+
+    if (called || mixerResult.initialized) {
+      logger.info(`Generating new mixer`);
+      logger.error(`Mixer already initialize`);
```
drewstone, line 49: This seems odd to have one after the other no? It doesn't look like it generates a new mixer but just returns undefined and so the log is a bit confusing.

drewstone: Might be good to remove the `logger.info` call.

drewstone (changes requested):
This looks great! And it actually fixes the graceful failing which was in the other PR, we could consider closing that one and just using this one?

drewstone:
We can merge this after spelling fixes and deal with hooking up logic later.

packages/page-mixer/src/components/deposit/DepositConsole.tsx
drewstone: We can deal with this in future PR.

PR #39, opened 2021-02-17
Add useFetch hook and present IP to user

### LINKED ISSUE #5: Display IP Address of user clearly
This is a trigger for people to turn on their VPNs

drewstone:
Resolves #5 and adds useFetch for async http calls.

### DIFF
--- a/packages/react-components/src/layout/index.tsx
+++ b/packages/react-components/src/layout/index.tsx
@@ -28,7 +28,7 @@ interface MainLayoutProps {
 const Main: FC<MainLayoutProps> = memo(({ children, enableCollapse = true, sidebar }) => {
   const { t } = useTranslation('react-components');
   const { init } = useApi();
-  const result = useFetch('https://ipapi.co/json/');
+  const result = useFetch('https://api.myip.com');
   const { allEndpoints, endpoint } = useSetting();
   const screen = useBreakpoint();
   const isAppReady = useIsAppReady();
--- a/packages/react-components/src/layout/index.tsx
+++ b/packages/react-components/src/layout/index.tsx
@@ -68,7 +68,7 @@ const Main: FC<MainLayoutProps> = memo(({ children, enableCollapse = true, sideb
     return (
       <Page>
         {result && (
-          <CAlert message={`${t('IP_ADDRESS')} ${result['city']}, ${result['region']} with IP: ${result['ip']}`} />
+          <CAlert message={`${t('IP_ADDRESS')} ${result['country']}, with IP: ${result['ip']}`} />
         )}
         <Page.Title
           breadcrumb={breadcrumb}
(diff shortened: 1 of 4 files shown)


drewstone:
Looks like:
image

drewstone:
image

drewstone:
@AhmedKorim wdyt about this?

packages/react-components/src/layout/index.tsx
```diff
@@ -28,6 +28,7 @@ interface MainLayoutProps {
 const Main: FC<MainLayoutProps> = memo(({ children, enableCollapse = true, sidebar }) => {
   const { t } = useTranslation('react-components');
   const { init } = useApi();
+  const result = useFetch('https://ipapi.co/json/');
```
shekohex, line 31: I would suggest using https://api.myip.com instead since this one is not working correctly.
```suggestion
  const result = useFetch('https://api.myip.com');
```

packages/react-components/src/layout/index.tsx
```diff
@@ -66,7 +67,9 @@ const Main: FC<MainLayoutProps> = memo(({ children, enableCollapse = true, sideb
 
     return (
       <Page>
-        {/* <CAlert message={t('TESTNET_SCAMES')} /> */}
+        {result && (
+          <CAlert message={`${t('IP_ADDRESS')} ${result['city']}, ${result['region']} with IP: ${result['ip']}`} />
```
shekohex, line 71:
```suggestion
          <CAlert message={`${t('IP_ADDRESS')} ${result['country']}, with IP: ${result['ip']}`} />
```

shekohex (changes requested):
Everything is good, Maybe @AhmedKorim has any other comments too on this.

packages/react-hooks/src/useFetch.ts
```diff
@@ -0,0 +1,22 @@
...
+      const response = await fetch(url);
+
+      console.log(response);
+      const json = await response.json();
+
+      setData(json);
+    }
+
+    fetchData();
+  }, [url]);
```
AhmedKorim, line 19: If the data is fetched after the component is unmounted  `setData` will throw an error as a state of the unmounted component can't be changed.

Also, It will help to use our `LoggerService` for future integrations

```suggestion
useEffect(() => {
    const controller = new AbortController();
    const logger = LoggerService.get('App');

    async function fetchData() {
      try {
        const response = await fetch(url, { signal: controller.signal });
        const json = await response.json();
        logger.trace(`Got date for url ${url}`, json);
        setData(json);
      } catch (e) {
        logger.error(e);
      }
    }

    fetchData();
    // Aborting the Request if the component is unmounted before Resolving
    return () => controller.abort();
  }, [url]);
```

AhmedKorim (changes requested):

PR #46, opened 2021-02-20
Add merkle group hook, pause mixer init, fix deposit dropdown and amounts

drewstone:


packages/page-mixer/src/components/deposit/DepositConsole.tsx
```diff
@@ -1,36 +1,25 @@
...
+import { Token, token2CurrencyId } from '@webb-tools/sdk-core';
 import { CurrencyId } from '@webb-tools/types/interfaces';
 import React, { FC, useCallback, useMemo, useState } from 'react';
 
 import { CardRoot, CardSubTitle, CardTitle, CTxButton, DepositTitle } from '../common';
 
 export const DepositConsole: FC = () => {
   const { api } = useApi();
-  const [token, setToken, { error: tokenError, setValidator: setTokenValidator }] = useInputValue<BalanceInputValue>({
+  const { mixerInfos } = useMixer();
```
AhmedKorim, line 15: I can see an issue with the current structure for `useMixer` it should be used inside the provider for the mixer-context, not directly.

we may move the code from the `useMixer` to the provider directly, or continue with the current and instead call `useMixerProvider`.



```suggestion
  const  mixerInfos  = useMixerInfos();
```

packages/page-mixer/src/components/deposit/DepositConsole.tsx
```diff
@@ -1,36 +1,25 @@
-import { BalanceInputValue, eliminateGap } from '@webb-dapp/react-components';
+import { BalanceInputValue } from '@webb-dapp/react-components';
 import AmountInput from '@webb-dapp/react-components/AmountInput/AmountInput';
 import { TokenInput } from '@webb-dapp/react-components/TokenInput';
-import { useApi, useBalance, useBalanceValidator, useConstants } from '@webb-dapp/react-hooks';
+import { useApi, useConstants, useMixer } from '@webb-dapp/react-hooks';
```
AhmedKorim, line 4: use `useMixerProvider` instead of `useMixer` which will point to the value from the context,
`useMixer` will return a new mixer ignoring the context

The context for the mixer doesn't contain the `mixerInfos`

```suggestion
import { useApi, useConstants, useMixerProvider , useMixerInfos} from '@webb-dapp/react-hooks';
```

packages/react-hooks/src/mixer/useMixer.ts
```diff
@@ -28,10 +30,19 @@ export const useMixerGroups = (): MixerAssetGroup[] => {
...
+    return mixerGroups || [];
+  }, [mixerGroups]);
+};
+
 const logger = LoggerService.new('MixerUsage');
 
 export const useMixer = () => {
-  const mixerAssetGroup = useMixerGroups();
+  const mixerIds = useMixerGroups();
+  const mixerInfos = useMixerInfos();
```
AhmedKorim, line 46: I'm not sure if we need `mixerInfos` for the mixer hook, `useMixer` should contain the code required for initing the mixer.

`useMixerInfos` can go to the context if caching is needed as it will be created one time, or it can be called directly.

AhmedKorim (changes requested):
@drewstone, Please check the review for `useMixerInfo` whether we need it to be cached or not

packages/react-hooks/src/mixer/useMixer.ts
drewstone: So mixer infos contains the details about the amounts each mixer requires. So we need it to display the mixer selections (by amounts).

PR #63, opened 2021-03-04
Added material-ui components

AhmedKorim:


packages/page-mixer/src/components/deposit/DepositConsole.tsx
```diff
@@ -123,7 +125,16 @@ export const DepositConsole: FC = () => {
...
+          <TriggerDeposit
+            onClick={() => {
+              setShowDepositModal(true);
+            }}
+            size='large'
+          >
+            Deposit
+          </TriggerDeposit>
+
+          {/*          <CTxButton
```
shekohex, line 137: Could you remove this commented code? since there is no need for it for now?

shekohex (changes requested):

PR #76, opened 2021-03-11
refactor withdraw and hooks

AhmedKorim:


packages/react-hooks/src/merkle/useMerkleGroups.ts
```diff
@@ -0,0 +1,32 @@
...
+    return this.inner?.toHuman().root_hash ?? null;
+  }
+
+  get rootHashU8a(): Uint8Array | null {
+    const rootHash = this.rootHash;
+    return rootHash ? hexToU8a(rootHash as any) : null;
+  }
+}
+
+export const useMerkleGroups = (id?: string | undefined) => {
```
shekohex, line 27: I guess the ID must be required, as we already querying for the group tree, right?
```suggestion
export const useGroupTree = (id: number): GroupTreeWrapper => {
```

packages/react-hooks/src/merkle/useMerkleGroups.ts
```diff
@@ -0,0 +1,32 @@
+import { useCall } from '@webb-dapp/react-hooks';
+import { GroupTree } from '@webb-tools/types/interfaces';
+import { useMemo } from 'react';
+import { hexToU8a } from '@polkadot/util';
+
+class MerkleGroupWrapper {
```
shekohex, line 6: I suggest we stick with the same name + `Wrapper` so this should be just `GroupTreeWrapper` so we don't make any other naming issue that we are doing this huge refactor to solve it :D
```suggestion
class GroupTreeWrapper {
```

shekohex (changes requested):

shekohex:
I suggest holding off this PR for a bit until the refactoring from webb.js is done.

packages/react-hooks/src/merkle/useMerkleGroups.ts
AhmedKorim: No, it's not required the code handles that case

packages/react-hooks/src/merkle/useMerkleGroups.ts
AhmedKorim: There is a Merkle group and mixer group, this naming follows the query path, not the type name

packages/react-hooks/src/merkle/useMerkleGroups.ts
```diff
@@ -0,0 +1,32 @@
...
+  get inner() {
+    return this._inner;
+  }
+
+  get ready() {
+    return Boolean(this.inner);
+  }
+
+  get rootHash() {
+    return this.inner?.toHuman().root_hash ?? null;
```
drewstone, line 18: can we avoid this pattern of using `toHuman`?

packages/react-hooks/src/merkle/useMerkleGroups.ts
drewstone: I agree with the renaming. If the API call is returning a GroupTree, then the hook should return that type + Wrapper.

packages/react-hooks/src/merkle/useMerkleGroups.ts
drewstone: I would suggest following the Type + Wrapper pattern not the query path pattern since that may update.

packages/react-hooks/src/mixer/useMixerGroupIds.ts
```diff
@@ -0,0 +1,29 @@
+import { useCall } from '@webb-dapp/react-hooks';
+import { MixerAssetGroup } from '@webb-tools/sdk-mixer';
+import { GroupId } from '@webb-tools/types/interfaces';
+import { useMemo } from 'react';
+
+class MixerGroupIdsWrapper {
```
drewstone, line 6: I would make this GroupIdWrapper

packages/react-hooks/src/mixer/useMixerGroupIds.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  // todo fix TokenSymbol type
+  // @ts-ignore
+  assetGroups(tokenSymbol = 'EDG', treeDepth = 32): MixerAssetGroup[] {
+    return this.inner.map((id) => new MixerAssetGroup(Number(id.toString()), tokenSymbol as any, treeDepth));
+  }
+}
+
+export const useMixerGroupIds = () => {
```
drewstone, line 24: While I know this is being called from mixer pallet, if the return type is GroupId, then we should still unify the wrapper around that type. This way we can reuse this between both pallets without confusion.

packages/react-hooks/src/mixer/useMixerGroups.ts
```diff
@@ -0,0 +1,30 @@
+import { useCall } from '@webb-dapp/react-hooks';
+import { MixerInfo, ScalarData } from '@webb-tools/types/interfaces';
+import { useMemo } from 'react';
+
+class MixerGroupWrapper {
```
drewstone, line 5: Lets use Type + Wrapper

drewstone (changes requested):

packages/react-hooks/src/withdraw/useWithdraw.tsx
```diff
@@ -1,63 +1,55 @@
...
-  const rootHash = useMemo<string | undefined>(() => groupTree?.toHuman().root_hash as any, [groupTree]);
-  const group = useCall<MixerInfo>(
-    'query.mixer.mixerGroups',
-    [noteMixerGroupId],
-    undefined,
-    undefined,
-    () => noteMixerGroupId !== undefined
-  );
+  const merkleGroups = useGroupTree(noteMixerGroupId?.toString());
+  const mixerGroups = useMixerInfo(noteMixerGroupId?.toString());
```
drewstone, line 30:
```suggestion
  const mixerInfo = useMixerInfo(noteMixerGroupId?.toString());
```

packages/page-mixer/src/components/deposit/DepositConsole.tsx
```diff
@@ -14,7 +14,7 @@ import { CardRoot, CardSubTitle, CardTitle, DepositTitle, TriggerDeposit } from
 const depositLogger = LoggerService.get('Deposit');
 // todo get this by selecting assets
 export const DepositConsole: FC = () => {
-  const mixerInfos = useMixerInfos();
+  const mixerInfos = useMixerInfo();
```
drewstone:
```suggestion
  const mixerInfo = useMixerInfo();
```

packages/react-hooks/src/mixer/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './useMixer';
 export * from './mixer-logger';
-export * from './useMixerGroups';
-export * from './useMixerInfos';
+export * from './useMixerGroupIds';
+export * from './useMixerGroupsEntries';
```
drewstone, line 4:
```suggestion
export * from './useMixerGroupEntries';
```

packages/react-hooks/src/mixer/useMixer.ts
```diff
@@ -1,40 +1,28 @@
...
-  }, [mixerGroupIds]);
-};
 
 const logger = LoggerService.new('MixerUsage');
 
 export const useMixer = () => {
-  const mixerIds = useMixerGroups();
-  const mixerInfos = useMixerInfos();
+  const mixerIds = useMixerGroupIds();
+  const mixerInfos = useMixerInfo();
```
drewstone, line 25:
```suggestion
  const mixerInfo = useMixerInfo();
```

packages/react-hooks/src/merkle/useGroupTree.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+}
+
+/**
+ * UseGroupTrees
+ *  @description   This will issue an RPC call to query.merkle.groups
+ *  @param {string} id which is the GroupId is optional if the is undefined the underlying rpc call won't take place
+ *  @return {GroupTreeWrapper}
+ * */
+export const useGroupTree = (id?: string | undefined) => {
```
drewstone, line 47: Specify return type

packages/page-mixer/src/components/deposit/DepositConsole.tsx
```diff
@@ -14,7 +14,7 @@ import { CardRoot, CardSubTitle, CardTitle, DepositTitle, TriggerDeposit } from
 const depositLogger = LoggerService.get('Deposit');
 // todo get this by selecting assets
 export const DepositConsole: FC = () => {
-  const mixerInfos = useMixerInfos();
+  const mixerInfos = useMixerGroupsEntries();
```
drewstone, line 17: the naming here is now inconsistent, please also add types to variables so its clear to you as well as the reader.

packages/react-hooks/src/mixer/useMixerInfo.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * UseMixerInfos
+ *  @description   This will issue an RPC call to query.mixer.mixerGroups
+ *   @param {string} id which is the GroupId is optional if the is undefined the underlying rpc call won't take place
+ *
+ *
+ *  @return {MixerInfoWrapper}
+ * */
+export const useMixerInfo = (id?: string | undefined) => {
```
drewstone, line 39: return type missing

drewstone (changes requested):

packages/react-hooks/src/mixer/useMixerInfo.ts
AhmedKorim: This will be tackled on another PR since we are changing the MixerContext api

packages/react-hooks/src/mixer/useMixer.ts
```diff
@@ -96,7 +84,7 @@ export const useMixer = () => {
     try {
       let gens = await generateBulletproof();
       logger.trace(`Bulletproof `, gens.length);
-      const mixer = await Mixer.init(new Worker(), mixerIds, gens);
+      const mixer = await Mixer.init(new Worker(), mixerIds.assetGroups(), gens);
```
drewstone, line 87: Are we intending to initialize the mixer with all the groups? This seems contrary to what we discussed a few days back as far as initializing ONLY one asset group (currency, amount) pair.

Since we are polling for only 1 mixer info, shouldn't we also reflect initializing the mixer with only 1 group as well?

packages/react-hooks/src/merkle/useGroupTree.ts
AhmedKorim: it's not really needed, however, let's make implicit

packages/react-hooks/src/mixer/useMixer.ts
AhmedKorim: Yeah this will be in a next pr

PR #104, opened 2021-06-18
Add beresheet mixers and successful deposit

nepoche:
Changes in this branch include various spelling fixes, and change of identifying chains by their chainId instead of networkType.

The current state of this branch allows for successful deposit on beresheet, but withdraw does not work.  I suspect this is due to a timeout on getDepositEvents, as we are trying to fetch all deposit events from block 0.  I think we will need to store all deposit events as they happen into a database for better retrieval when a user wants to withdraw.


packages/contracts/deployedContracts.js
```diff
@@ -0,0 +1,28 @@
+
```
nepoche, line 1: Delete this file

AhmedKorim:
Looks good, will test this on Monday

packages/react-environment/src/WebbProvider.tsx
```diff
@@ -106,20 +113,24 @@ export const WebbProvider: FC<WebbProviderProps> = ({ applicationName = 'Webb Da
...
-        switch (chainType) {
-          case WebbEVMChain.Main:
-            storage = mainS;
-            break;
-          case WebbEVMChain.Rinkeby:
-            storage = rainkybeS;
-            break;
-        }
-        const webbWeb3Provider = await WebbWeb3Provider.init(web3Provider, rainkybeS);
+          switch (net) {
```
AhmedKorim, line 119: It's good to init storages lazily, hover other storages are needed for the provider if the users decided to change the network.

packages/react-environment/src/WebbProvider.tsx
```diff
@@ -106,20 +113,24 @@ export const WebbProvider: FC<WebbProviderProps> = ({ applicationName = 'Webb Da
...
         let storage: Storage<EVMStorage>;
-        switch (chainType) {
-          case WebbEVMChain.Main:
-            storage = mainS;
-            break;
-          case WebbEVMChain.Rinkeby:
-            storage = rainkybeS;
-            break;
-        }
-        const webbWeb3Provider = await WebbWeb3Provider.init(web3Provider, rainkybeS);
```
AhmedKorim, line 122: It's good we init the storages lazily, however, they are needed if the user decided to switch the network from the extension, then there should be a way to swap the storage

AhmedKorim: @nepoche  a solution for this would be to extend the EventBus for `Web3Provider` and emit an event `netowrkChanged` and pass the instance for web3Provider and then handle swapping the storage

nepoche: What about instead we refresh the application upon extension network switch as the metamask docs suggest? Then the storage would be handled by the `WebbProvider` init. I believe page refresh is how most DApps that I have used work.

drewstone:
Some merge conflicts left @nepoche

nepoche:
Fixed merge conflict and squashed for clean commit history

AhmedKorim:
Looks good

PR #106, opened 2021-06-22
Add fields for syncedBlock and leaves in EVMChainStorage class

nepoche:
I'm trying to determine:
1. Where should we store the dynamic data of syncedBlock and leaves in the localStorage? 

- I have added it to EVMChainStorage in this PR.

2. Where should we update the data for syncedBlock and leaves? 
 
- I have noted the place in the contract and comments for how we can update this storage.


AhmedKorim:
@nepoche are we mergin this or it was just a plaground?

drewstone:
We can use these places to discuss both options IMO. If we agree we like
the proposal (and any changes) then it’s worth merging. I’ll aim to review
this tomorrow.

drewstone:
I suggest renaming these files too tbh. `evm-chain-storage.interface.ts` is a mouthful. What's the point of naming things like this?

Also, we're mixing camel case named files `WebbProvider`, `SettingsManager`, with hyphen delineated files. Lets stick to one...

I vote camelcasing. My suggestion is:
- Create a folder `packages/apps/src/configs/storages/evm/`
- Name the chain storage interface `evm/IChainStorage.ts` or `evm/IStorage.ts`
- Name the storage implementation `evm/storage.ts`.

packages/contracts/src/contracts/anchor.ts
```diff
@@ -48,8 +54,9 @@ export class AnchorContract {
     const overrides = {
       gasLimit: 6000000,
       gasPrice: utils.toWei('1', 'gwei'),
-      value: '0x16345785D8A0000',
+      value: await this.denomination,
```
drewstone, line 57: Maybe I don't know much about getters but why are we awaiting this?

packages/apps/src/configs/storages/evm-storage.ts
```diff
@@ -0,0 +1,93 @@
...
+  {
+    size: 10000,
+    address: '0x626FEc5Ffa7Bf1EE8CEd7daBdE545630473E3ABb',
+    symbol: 'tEDG',
+    syncedBlock: 1, 
+    leaves: [],
+  },
+]);
+
+const defaultHandler: Omit<StorageHandler<EVMStorage>, 'inner'> = {
```
drewstone, line 56: Any reason why we don't call these methods `get` and `set`?

packages/apps/src/configs/storages/evm-storage.ts
```diff
@@ -0,0 +1,93 @@
...
+  async fetch(key: string) {
+    const data = localStorage.getItem(key);
+    if (!data) {
+      return {
+        nativeAnchor: new EvmChainStorage([]),
+      };
+    }
+    const address = JSON.parse(data).contractsAddresses;
+    return {
+      nativeAnchor: new EvmChainStorage(address),
```
drewstone, line 66: Is this currently a nativeAnchor because the only mixers we currently have supported in the dApp are native? I suppose we'll want to plan accordingly for ERC20 supported mixers. Might be worth updating this now, thoughts?

packages/apps/src/configs/storages/evm-storage.ts
```diff
@@ -0,0 +1,93 @@
+import { EvmChainStorage } from '@webb-dapp/apps/configs/storages/evm-chain-storage.interface';
+import { StorageHandler } from '@webb-dapp/utils';
+import { EVMStorage } from '@webb-dapp/react-environment/api-providers/web3';
+
+const rinkebyStore = new EvmChainStorage([
+  {
+    size: 0.1,
+    address: '0x876eCe69618e8E8dd743250B036785813824D2D7',
+    symbol: 'ETH',
+    syncedBlock: 120000, // should be hardcoded to deployed block number
```
drewstone, line 10: I still don't know what syncedBlock means. This should err on the side of being as verbose as needed. Is this the latest synced block? If so, lets call it, `latestSyncedBlock` or `lastSyncedBlock` or `lastQueriedBlock`, etc.

I think we should also add another parameter for `createdAtBlock` to indicate the block at which these mixers were created.  This way, for new clients, we can begin fetching events immediately from the point they were created rather than before ever, mistakenly

packages/apps/src/configs/storages/evm-storage.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const defaultHandler: Omit<StorageHandler<EVMStorage>, 'inner'> = {
+  async fetch(key: string) {
+    const data = localStorage.getItem(key);
+    if (!data) {
+      return {
+        nativeAnchor: new EvmChainStorage([]),
+      };
+    }
+    const address = JSON.parse(data).contractsAddresses;
```
drewstone, line 64: This doesn't seem right. The return type is called address but the parameter is called `contractsAddresses`? I'm not sure I understand how you're storing the data in localStorage. This ought to be updated and more clearly defined right?

Do share where you think we will actually `commit` the data. I do recommend discussing this and/or being opinionated and deciding with confidence.

packages/contracts/src/contracts/anchor.ts
nepoche: This makes a call to the RPC to read the denomination from the contract deployed on the evm. We need to wait for the RPC response.

packages/apps/src/configs/storages/evm-chain-storage.interface.ts
```diff
@@ -0,0 +1,22 @@
+import { MixerSize } from '@webb-dapp/react-environment';
+
+export class EvmChainStorage {
```
nepoche, line 3: I think this class should be renamed to `EvmChainMixerInfo`.  This is because most of the fields inside of this class should not be dealing with storage.

packages/apps/src/configs/storages/evm-storage.ts
nepoche: I agree the field can be named something more descriptive like `lastQueriedBlock`.

I think that these `EvmChainStorage` objects will be passed default values as params to the constructor. Therefore, the `lastQueriedBlock` can pass the deployed block number, and new clients will only ever start querying from the deployed block.  In the constructor, we can read from the localStorage to determine if the client has anything in storage for the particular mixer.

packages/contracts/src/contracts/anchor.ts
drewstone: Awesome, so this is mostly to generalise over Substrate chains and EVM, since we get denomination info often asynchronously. Good stuff!

PR #107, opened 2021-06-22
Added Currency config and implemented  the token selector

AhmedKorim:


packages/apps/src/configs/wallets/currency-config.ts
```diff
@@ -0,0 +1,31 @@
...
+import { AppConfig } from '@webb-dapp/react-environment';
+import React from 'react';
+
+export const currenciesConfig: AppConfig['currencies'] = {
+  1: {
+    name: 'Edgware token',
+    symbol: 'EDG',
+    color: '',
+    id: 1,
+    supportingChainsIds: [3, 1],
```
nepoche, line 12: Chains already know the IDs of currencies, why do currencies need to know the chains which support them?

AhmedKorim: for easier access, we can just do this on the class for `Currency` too , from chains config we get the supported chains from a currency

AhmedKorim: YOu have a good point BTW will remove it In the meantime and add implementation for this on the Currency

PR #109, opened 2021-06-24
Better Error handler

AhmedKorim:


packages/apps/src/App.tsx
```diff
@@ -13,11 +13,13 @@ import { config as routerConfig } from './router-config';
 
 const appLogger = LoggerService.new('App');
 const App: FC = () => {
+  const {} = useWebContext();
```
drewstone, line 16: This seems out of place?

packages/apps/src/configs/wallets/chain-config.ts
```diff
@@ -1,13 +1,15 @@
...
 import EtherLogo from '@webb-dapp/apps/configs/wallets/logos/Eth';
+import { WebbEVMChain } from '@webb-dapp/react-environment/api-providers/web3';
 
 export const chainsConfig: AppConfig['chains'] = {
   // edgware
   1: {
     id: 1,
     group: 'edgware',
     tag: 'dev',
+    evmId: undefined,
```
drewstone, line 12: If we use enums then we should make sure:
- Edgeware mainnet is 2021
- Edgeware beresheet is 2022

packages/react-environment/src/api-providers/polkadot/webb-polkadot-provider.ts
```diff
@@ -24,10 +37,38 @@ export class WebbPolkadot extends PolkadotProvider implements WebbApiProvider<We
     };
   }
 
-  static async init(appName: string, endpoints: string[]): Promise<WebbPolkadot> {
-    const [apiPromise, injectedExtension] = await PolkadotProvider.getParams(appName, endpoints);
+  async awaitMetaDataCheck() {
+    /// delay some time till the UI is instantiated and then check if the dApp needs to update extension meta data
+    await new Promise((r) => setTimeout(r, 3000));
```
drewstone, line 42: This isn't great practice, can we remove this? What errors occur when you do?

packages/wallet/src/providers/polkadot/polkadot-provider.ts
```diff
@@ -48,12 +58,52 @@ export class PolkadotProvider extends EventBus<ExtensionProviderEvents> {
...
+      });
+      let reported = false;
+      wsProvider.on('error', (e) => {
+        if (reported) {
+          return;
+        }
+        let interActiveFeedback: InterActiveFeedback;
+        const body = InterActiveFeedback.feedbackEntries([
+          {
+            header: 'Failed to establish WS connection',
```
drewstone, line 74: we should probably abstract all our error messages into a common folder/location as well and use Enums.

drewstone (changes requested):
Thanks for adding comments to the code, it's really helpful!

The design of this seems great and well-intentioned. I am not sure about adding hardcoded delays into the code as that does seem like something we should avoid. How and why are we not able to wait until certain processes are loaded? A 3 second delay seems arbitrary and could provide issues in the future.

It seems we're just using this error handler in api init correct? Just want to make sure I'm forming a good mental model. The plan is to use this everywhere we have errors right?

packages/apps/src/App.tsx
AhmedKorim: The`InteractiveFeedback`is going to be used with errors/info/warning that requires feedback from the user and trigged view via the providers, not just errors, and this is meant to be decoupled from the view.

`notificationApi` is used with this for showing toast messages without feedback although `notificationApi` isn't really decoupled and relay on some react logic not sure if it needs to be decoupled(fully)  as well at least ATM

drewstone: I just mean the return type is `{}`. Is something meant to be returned here?

packages/wallet/src/providers/polkadot/polkadot-provider.ts
AhmedKorim: Added implementation for the Error here https://github.com/webb-tools/webb-dapp/blob/c936d7541f44e5216b6f833790710993deb39843/packages/react-components/src/InteractiveFeedbackView/InteractiveErrorView.tsx#,
and as you suggested we can handle the  content for error messages

packages/apps/src/App.tsx
AhmedKorim: This is a dead code removed it on future commits

packages/react-environment/src/WebbProvider.tsx
```diff
@@ -51,31 +117,213 @@ export const WebbProvider: FC<WebbProviderProps> = ({ applicationName = 'Webb Da
...
+            const webbWeb3Provider = await WebbWeb3Provider.init(web3Provider, storage);
+            const accounts = await webbWeb3Provider.accounts.accounts();
+            setAccounts(accounts);
+            await setActiveAccount(accounts[0]);
+            await webbWeb3Provider.accounts.setActiveAccount(accounts[0]);
+            setActiveApi(webbWeb3Provider);
+            webbWeb3Provider.on('providerUpdate', async ([chainId]) => {
+              const nextChain = Object.values(chains).find((chain) => chain.evmId === chainId);
+              if (!nextChain) {
+                try {
```
nepoche, line 189: Should we be trying anything if we do not support the evm chain id?

AhmedKorim: this call `WebbWeb3Provider.storageName(chainId)` will throw `UnsupportedChain` Error catching this and show an interactiveFeedback

packages/apps/src/configs/wallets/chain-config.ts
AhmedKorim: Will handle this on a refactor PR after the #111

PR #110, opened 2021-06-24
Add Browser Storage for Leaves of Mixers

nepoche:
Things that were attempted in this PR:

1. `EvmChainStorage` Renamed to `EvmChainMixersInfo`
- Implement any methods that would require getting information about a mixer from this class. And setup information queries based off token Symbol to support multiple currencies.
- Inside of this class, I tried to split the information that would be accessible between `StaticMixerInfo` and `DynamicMixerInfoStore`. This is because I wanted the localStorage to be the one truth of state for lastQueriedBlock as well as the fetched leaves. 

2. Accessing the storage from the `AnchorContract`
- This seems like the most natural place to retrieve and update the leaves. As it is right now, it looks directly at the global localStorage. Should this be changed to pass in an instance of `EvmChainMixersInfo`?
- There are issues with retrieving the storage if it has not been initialized.


packages/apps/src/configs/storages/evm-chain-storage.interface.ts
```diff
@@ -1,9 +1,35 @@
...
+    
+    public contractsInfo: { 
+      address: string; 
+      size: number; 
+      symbol: string; 
+      syncedBlock: number;
+      leaves: string[];
+    }[]
+  ) {
+    // On construction, retrieve the
```
drewstone, line 28: Unfinished or can be removed?

packages/apps/src/configs/storages/evm-chain-storage.interface.ts
```diff
@@ -1,9 +1,35 @@
...
+      syncedBlock: number;
+      leaves: string[];
+    }[]
+  ) {
+    // On construction, retrieve the 
 
-  get mixerSize(): MixerSize[] {
+  }
+
+  getMixersInfo(): MixerSize[] {
```
drewstone, line 32: This still seems better named as `getMixerSizes` given the return type.

packages/apps/src/configs/storages/evm-storage.ts
```diff
@@ -1,81 +1,44 @@
...
-import { StorageHandler } from '@webb-dapp/utils';
-import { EVMStorage } from '@webb-dapp/react-environment/api-providers/web3';
-
-const rinkebyStore = new EvmChainStorage([
+export const rinkebyMixers = [
   {
     size: 0.1,
     address: '0x876eCe69618e8E8dd743250B036785813824D2D7',
     symbol: 'ETH',
+    createdAtBlock: 120000, // should be hardcoded to deployed block number
```
drewstone, line 6: 👍

packages/apps/src/configs/wallets/chain-config.ts
```diff
@@ -28,11 +28,11 @@ export const chainsConfig: AppConfig['chains'] = {
     logo: EdgwareLogo,
     currencies: [
       {
-        currencyId: 1,
+        currencyId: 2,
         enabled: true,
       },
     ],
-    nativeCurrencyId: 1,
+    nativeCurrencyId: 2,
```
drewstone, line 35: let me know if you're not understanding these values. For background, the native currency will likely always be 0.

packages/apps/src/configs/wallets/chain-config.ts
```diff
@@ -28,11 +28,11 @@ export const chainsConfig: AppConfig['chains'] = {
     logo: EdgwareLogo,
     currencies: [
       {
-        currencyId: 1,
+        currencyId: 2,
```
drewstone, line 31: CurrencyIDs on Substrate are identifiers for non-native tokens. We reserve one for the native token to not interfere and otherwise choose higher values (though we'll see how other chains handle this in time).

packages/contracts/src/contracts/anchor.ts
```diff
@@ -17,7 +21,7 @@ export class AnchorContract {
   private _contract: Anchor;
   private readonly signer: Signer;
 
-  constructor(private web3Provider: providers.Web3Provider, address: string) {
+  constructor(private mixersInfo: EvmChainMixersInfo, private web3Provider: providers.Web3Provider, address: string) {
```
drewstone, line 24: can we opt for `EvmChainMixerInfos`? and `mixerInfos`?

packages/contracts/src/contracts/anchor.ts
```diff
@@ -65,56 +69,57 @@ export class AnchorContract {
...
 
       // If there is a timeout, query the logs in block increments.
       if (e.code == -32603) {
-        for (var i=startingBlock; i < currentBlock; i+= 1000)
+        for (var i=startingBlock; i < currentBlock; i+= 300)
         {
           logs = [...logs, ...(await this.web3Provider.getLogs({
             fromBlock: i,
-            toBlock: (currentBlock - i > 1000) ? i + 1000 : currentBlock,
+            toBlock: (currentBlock - i > 300) ? i + 300 : currentBlock,
```
drewstone, line 97:
```suggestion
            toBlock: (currentBlock - i > 512) ? i + 512 : currentBlock,
```

packages/contracts/src/contracts/anchor.ts
```diff
@@ -65,56 +69,57 @@ export class AnchorContract {
...
+        toBlock: currentBlock,
         ...filter,
       });
     } catch (e) {
       console.log(e);
 
       // If there is a timeout, query the logs in block increments.
       if (e.code == -32603) {
-        for (var i=startingBlock; i < currentBlock; i+= 1000)
+        for (var i=startingBlock; i < currentBlock; i+= 300)
```
drewstone, line 93:
```suggestion
        for (var i=startingBlock; i < currentBlock; i+= 512)
```

packages/react-environment/src/api-providers/polkadot/polkadot-mixer-withdraw.ts
```diff
@@ -81,9 +80,19 @@ export class PolkadotMixerWithdraw extends MixerWithdraw<WebbPolkadot> {
...
-
+    const storage = this.inner.storage;
+    const mixerGroup = (await storage.get(String(treeId))) || {
+      0: { leaves: [] },
+    };
+    const mixerGroupForChain = mixerGroup[0];
+    const leaves: Uint8Array[] = mixerGroupForChain.leaves.map((leafStr) => {
+      return hexToU8a(leafStr);
+    });
+    // todo check this if the index starts from 1, or zero
```
drewstone, line 91:
```suggestion
    // TODO: check this if the index starts from 1, or zero
```

packages/react-environment/src/api-providers/polkadot/polkadot-mixer-withdraw.ts
```diff
@@ -81,9 +80,19 @@ export class PolkadotMixerWithdraw extends MixerWithdraw<WebbPolkadot> {
...
+    from = Math.max(leaves.length - 1, 0);
+    to = from + 511;
     while (done === false) {
+      console.log(`Fetching leaves for treeId${treeId} from :${from} to ${to}`);
```
drewstone, line 95: I would use the proper logger that we have, I think there should be examples somewhere of its usage right @AhmedKorim ?

packages/react-environment/src/api-providers/web3/EvmChainMixersInfo.tsx
```diff
@@ -0,0 +1,164 @@
...
+        title: `${contract.size} ${contract.symbol}`,
+      };
+    })
+  }
+
+  async getMixerInfoStorage(contractAddress: string) {
+    // create the mixerStorage if it didn't exist
+    if (!this.mixerStorage)
+    {
+      console.log('mixer storage is being created');
```
drewstone, line 123: Use proper logger if possible.

packages/react-environment/src/api-providers/web3/EvmChainMixersInfo.tsx
```diff
@@ -0,0 +1,164 @@
...
+      this.mixerStorage = await mixerStorageFactory(this.chainId);
+      console.log(`This is the mixerstorage: ${this.mixerStorage}`);
```
drewstone, line 125: Perhaps since these may be intentioned to be removed in the future, opt for a better log message and use the debug log setting.

packages/react-environment/src/api-providers/web3/web3-mixer-deposit.ts
```diff
@@ -48,8 +46,7 @@ export class Web3MixerDeposit extends MixerDeposit<WebbWeb3Provider, DepositPayl
     };
   }
 
-  async getSizes(contract: string = 'nativeAnchor'): Promise<MixerSize[]> {
-    const storage = await this.inner.chainStorage;
-    return storage[contract].mixerSize;
+  async getSizes(): Promise<MixerSize[]> {
+    return this.inner.getMixersSizes(WebbWeb3Provider.getNativeCurrencySymbol(await this.inner.getChainId()));
```
drewstone, line 50: I think you changed the name of this function above to `getMixersInfo` no? Nonetheless, we should call it `getMixerSizes` as was brought up at the function def.

packages/react-environment/src/api-providers/web3/web3-mixer-withdraw.ts
```diff
@@ -8,10 +8,12 @@ export class Web3MixerWithdraw extends MixerWithdraw<WebbWeb3Provider> {
...
   }
 
-  async withdraw(note: string, recipient: string, contractName: string = 'nativeAnchor'): Promise<void> {
+  async withdraw(note: string, recipient: string): Promise<void> {
     this.emit('stateChange', WithdrawState.GeneratingZk);
     const evmNote = EvmNote.deserialize(note);
-    const contract = await this.inner.getContract(evmNote.amount, contractName);
+    const contract = await this.inner.getContractBySize(evmNote.amount, evmNote.currency);
+    if (!contract)
+      throw new Error('Invalid note!');
```
drewstone, line 16: I'd add a TODO to add this to the dedicated error handler once it's ready.

drewstone (changes requested):
Looks good. For naming conventions I do not myself like <PLURAL><SINGULAR> naming when talking about many <SINGULAR> things. For this reason I think we should use:
- MixerInfos
- MixerSizes
- MixerStorages
rather than MixersXYZ. Otherwise there's some small changes I'd make to logging.

Overall this looks great! Would like to have @AhmedKorim give a pass too. Great work @nepoche

drewstone:
Also a better name for EvmChainStorage might actually be EvmMixerStorage rather than what we have opted for here. What do you think @nepoche ?

packages/react-environment/src/api-providers/web3/EvmChainMixersInfo.tsx
nepoche: I've removed them.

packages/react-environment/src/api-providers/web3/web3-mixer-deposit.ts
nepoche: I've renamed the functions and type to getMixerTitles and MixerTitle respectively.

nepoche:
> Also a better name for EvmChainStorage might actually be EvmMixerStorage rather than what we have opted for here. What do you think @nepoche ?

I prefer EvmChainStorage because we are storing information about all mixers under one key in local storage depending on the chain id.

packages/react-environment/src/api-providers/web3/web3-mixer-deposit.ts
drewstone: On the same topic, please remove MixersXYZs naming convention. The plural should not be applied twice. It should be MixerXYZs.

For mixer titles, it should be getMixerTitles not getMixersTitles. The latter is more awkward.

For example, for something more semantically clear, if I had an asset pallet where each asset had a name, I wouldn't want to name things as getAssetsNames, I'd want to say getAssetNames which is similar to getAllAssetNames. So semantically, we're looking for "get me all mixer sizes" not "get me all mixers sizes". Do you agree or disagree?

PR #111, opened 2021-06-28
Add support for EVM browser storage of mixer leaves

nepoche:
A few things need to happen before this PR can be merged:

1. Need to address / make note of extreme slowdown in `eth_getLogs` calls for frontier EVM. Upon first deployment,
2. Need to handle paths not resolving correctly when attempting to use `@webb-dapp`. Right now I have the code using relative paths with ..


packages/react-environment/src/api-providers/web3/EvmChainMixersInfo.ts
```diff
@@ -0,0 +1,93 @@
+import { MixerTitle } from '@webb-dapp/react-environment';
+import { Storage } from '@webb-dapp/utils';
+import { MixerStorage, evmChainStorageFactory } from '../../../../apps/src/configs/storages/EvmChainStorage';
```
nepoche, line 3: When attempting to import from `@webb-dapp/apps/configs/storages/EvmChainStorage`, the file cannot be found.

packages/contracts/src/contracts/anchor.ts
```diff
@@ -65,56 +63,58 @@ export class AnchorContract {
...
+        for (var i=startingBlock; i < currentBlock; i+= 50)
         {
           logs = [...logs, ...(await this.web3Provider.getLogs({
             fromBlock: i,
-            toBlock: (currentBlock - i > 1000) ? i + 1000 : currentBlock,
+            toBlock: (currentBlock - i > 50) ? i + 50 : currentBlock,
             ...filter,
           }))];
+
+          console.log(`Getting logs for block range: ${i} through ${i+50}`)
```
nepoche, line 94: This should be removed before integrating, but it shows the issue with EVM. This worked with an increment of 1000 around the time of beresheet v2 launch ~block 3000.

packages/react-environment/src/api-providers/web3/EvmChainMixersInfo.ts
drewstone: Hmm, @AhmedKorim any idea here?

packages/contracts/src/contracts/anchor.ts
drewstone: Can you try to use the dedicated logger? I thought we had one in he dApp. You can keep it in but set it as a debug log.

packages/react-environment/src/webb-context/mixer/mixer-deposit.ts
```diff
@@ -12,7 +12,7 @@ export type MixerDepositEvents = {
 };
 export type DespotStates = 'ideal' | 'generating-note' | 'depositing';
 
-export type MixerSize = {
+export type MixerTitle = {
```
AhmedKorim, line 15: `MixerSize` isn't just for display and more data will be added to it

drewstone (commented):
This looks great @nepoche, if @AhmedKorim thinks this is good we can merge too!

AhmedKorim (changes requested):
Looks good the `getNativeCurrencySymbol` helper just need to consume the chained config and currency config, plus some notes

packages/apps/src/configs/evm/SupportedMixers.ts
```diff
@@ -0,0 +1,78 @@
...
+    case WebbEVMChain.Main:
+      return 'ETH';
+    case WebbEVMChain.Beresheet:
+      return 'tEDG';
+    case WebbEVMChain.Edgeware:
+      return 'EDG';
+    default:
+      throw new Error('unsupported chain');
+  }
+}
```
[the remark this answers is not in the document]
nepoche: What would the code look like for this? 
Something like: 
```
    case WebbEVMChain.Rinkeby:
         return currenciesConfig[chainsConfig[3].nativeCurrencyId].symbol;
```

Since we do not have a clean mapping from the evm reported chainId and what we consider a chainId in the dapp's `chain-config.ts`, it seems this would require some hardcoding for the index of the chainConfig.

packages/contracts/src/contracts/anchor.ts
nepoche: I put this log in the mixerLogger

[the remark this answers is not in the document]
nepoche: Updated, good catch.

packages/react-environment/src/webb-context/mixer/mixer-deposit.ts
nepoche: Changed it back, but curious what we will be adding to this field.

packages/apps/src/configs/evm/SupportedMixers.ts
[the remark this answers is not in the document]
AhmedKorim: yeah this would be fine but let's handle this on another PR
Great work :D

packages/react-environment/src/webb-context/mixer/mixer-deposit.ts
AhmedKorim: Don't know in the meantime as the type has the id which can be used to find more related information

AhmedKorim:
Looks good so far, the supported mixers should be fetching the native currency from the config not hardcoded values, but let's handle this on another PR

PR #113, opened 2021-06-29
Harmony Testnet Shard1 Support with Metamask

nepoche:


packages/apps/src/configs/evm/SupportedMixers.ts
```diff
@@ -9,6 +10,15 @@ export const getNativeCurrencySymbol = (chainID: number): string => {
   return 'Unknown';
 };
 
+export const getEVMChainName = (chainID: number): string => {
```
nepoche, line 13: Should this function and other helper functions be moved to a different file?

AhmedKorim: we can keep them in the meantime

AhmedKorim (approved):

AhmedKorim:
Looks good!

PR #121, opened 2021-07-08
Use mixers which depend on the same keys used in tornado's trusted setup

nepoche:


drewstone (approved):
Looks good, does this work for all of the mixers?

AhmedKorim (approved):

AhmedKorim:
Looks good

PR #122, opened 2021-07-10
Wallet connect && mobile wallets fixes

AhmedKorim:


packages/react-environment/src/WebbProvider.tsx
```diff
@@ -221,9 +221,63 @@ export const WebbProvider: FC<WebbProviderProps> = ({ applicationName = 'Webb Da
...
+        case WalletId.WalletConnect:
           {
-            /// init provider from the extension
-            const web3Provider = await Web3Provider.fromExtension();
+            let web3Provider: Web3Provider;
+            if (wallet?.id === WalletId.WalletConnect) {
+              const provider = new WalletConnectProvider({
+                rpc: {
+                  //default on metamask
+                  [WebbEVMChain.EthereumMainNet]: 'https://mainnet.infura.io/v3/9aa3d95b3bc440fa88ea12eaa4456161',
```
nepoche, line 231: I think these infura endpoints should probably be behind a .env

drewstone: Yea, at least the IDs (although this one's already committed).

AhmedKorim: Those are coming from MetaMask Extention, we may add those to the configuration object as well

PR #123, opened 2021-07-12
Implement IP Display UI Component

nepoche:


drewstone:
Post pics!

nepoche:
Screen Shot 2021-07-12 at 2 21 00 PM
Screen Shot 2021-07-12 at 2 27 55 PM

AhmedKorim:
@nepoche  do you think we can wait for the Relayer integration and then update this pr to use it, or we can do it later.

Looks good so far, for the white theme can you reduce the shadow to be less than the Mixer component?

PR #126, opened 2021-07-19
QA: Fix homepage with chain issues

nepoche:
* Split out container display from the `InputLabel` component into an `InputSection` component. This was necessary for the 'Token for Deposit' label.

* Typography overrides for variants used in the AccountManager display.


nepoche:
Some Pics for your viewing:
Before:

BeforeHomepageWithChain


After:

AfterHomepageWithChain

drewstone (approved):
Looks great @nepoche !

PR #131, opened 2021-07-26
Fixes and improvements for the chain switcher component

nepoche:
- [ ]  Should show an error
- [x]  The component needs clear boundaries.
- [x]  Remove URLs from the list items.
- [x]  Remove connectable via
- [x]  Make the network name larger
- [x]  Add clear boundaries between list items (like Anyswap)
- [x]  Add radio buttons for "live", "test", "dev"
    - As a user, I can click to view all "live" chains and see only "live" production chains.
    - As a user, I can click "test" and see only the available testnet chains.
    - Should be improved for an 'all' or deselect.
- [ ]  Live should be default
    - Default is test,
- [x]  If you're on a "live"/"test"/"dev" chain already then it should show you those options
- [ ]  Search bar for filtering, which also respects the radio buttons
- [x]  Clicking outside of modal should close modal

### DIFF
--- a/packages/apps/src/configs/chains/chain-config.ts
+++ b/packages/apps/src/configs/chains/chain-config.ts
@@ -90,54 +90,6 @@ export const chainsConfig: AppConfig['chains'] = {
     ],
     nativeCurrencyId: WebbCurrencyId.ETH,
   },
-  // [ChainId.Ropsten]: {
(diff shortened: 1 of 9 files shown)


nepoche:
Pictures of changes to the chain switcher:

ChainSwitcherUpdated

PR #136, opened 2021-08-03
Remove chains which do not have deployments from view

nepoche:


packages/ui-components/src/NetworkManger/NetworkManager.tsx
```diff
@@ -108,7 +108,7 @@ export const NetworkManager: React.FC<NetworkManagerProps> = () => {
       <FilterSection>
         <FormControl>
           <RadioGroup value={radioButtonFilter} onChange={handleRadioFilter} row>
-            <FormControlLabel value="live" control={<Radio />} label="live"/>
+            {/* <FormControlLabel value="live" control={<Radio />} label="live"/> */}
```
drewstone, line 111: I'd consider leaving this and handling this more gracefully. Like, "if there are no live chains don't show the live chain button". It should be possible to handle it like:
```
liveChains.length > 0 && <FormControlLabel value="live" control={<Radio />} label="live"/>
```

drewstone (changes requested):

packages/ui-components/src/NetworkManger/NetworkManager.tsx
drewstone: what do you think @nepoche ?

nepoche: Fair argument, but I think the downside to that approach is that the components start to get messy when there's a lot of conditional rendering.

The downside to this approach is we need to remember to uncomment this radio-button filter when we eventually do some live deployments.

I'll let @AhmedKorim give his input too.

PR #139, opened 2021-08-23
Throw an unsupported network error if the chain is not recognized

nepoche:
These changes remove dependency on activeChain variable when initializing interactive feedback in the app.  This allows for an unsupported network error on page load.

It is a bit awkward though because the error will show the Ethereum Mainnet as something 'unknown'. Maybe it would be better to move forward with an update to WebbErrors to differentiate between 'UnknownChain' and 'UnsupportedChain'?

### DIFF
--- a/packages/apps/src/configs/evm/SupportedMixers.ts
+++ b/packages/apps/src/configs/evm/SupportedMixers.ts
@@ -41,9 +41,7 @@ export const rinkebyMixers: MixerInfo[] = [
   },
 ];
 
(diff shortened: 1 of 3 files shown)


drewstone:
@nepoche can you make sure to lint, looks good otherwise!

nepoche:
The linting has not been enforced in this repo for a while, can I make necessary linting updates as a separate PR to bring it up to speed?

drewstone:
Sure thing

PR #145, opened 2021-08-31
Fix multiple WebbWeb3Provider creation and multi-notification

nepoche:
Each time the switchChain function is called, a WebbWeb3Provider instance was being created. The setting of the eventListener prevented cleanup, but the unregistering of the event with 'off' would remove the providerUpdate listen functionality in error scenarios (essentially every scenario when switch is attempted via the UI). This change minimizes the WebbWeb3Providers that are created, to hopefully prevent multiple notifications for chain switch.


AhmedKorim:
Sounds good!

AhmedKorim (approved):

PR #147, opened 2021-08-31
Resolve a lot of style / lint errors

nepoche:


nepoche:
Can't figure out what this linting error is pointing to

drewstone:
if we are to merge a huge lint PR in we should make sure it at least fixes the lint error in CI or to fix the lint config itself so that it passes.

PR #148, opened 2021-09-01
Fix withdraw leaf duplications being stored in local storage

nepoche:


packages/contracts/src/contracts/anchor.ts
```diff
@@ -110,26 +107,46 @@ export class AnchorContract {
...
+   *  4- Compare against historical roots before adding to local storage
+   *  5- return the path to the leaf.
    * */
   async generateMerkleProof(deposit: Deposit) {
-    const leaves = await this.getDepositLeaves();
-    logger.trace(`Leaves ${leaves.length}`, leaves);
-    const tree = new MerkleTree('eth', 20, leaves);
-    let leafIndex = leaves.findIndex((commitment) => commitment == bufferToFixed(deposit.commitment));
+    const storedContractInfo = await this.mixersInfo.getMixerStorage(this._contract.address);
+    const tree = new MerkleTree('eth', 20, storedContractInfo.leaves);
```
drewstone, line 127: Can we query the contract for the merkle tree height as well? Currently 20 is hardcoded and should be swapped for a definitely foolproof value.

packages/contracts/src/contracts/anchor.ts
```diff
@@ -110,26 +107,46 @@ export class AnchorContract {
...
+
+    const fetchedLeaves = await this.getDepositLeaves(lastQueriedBlock + 1);
+    logger.trace(`New Leaves ${fetchedLeaves.newLeaves.length}`, fetchedLeaves.newLeaves);
+
+    tree.batch_insert(fetchedLeaves.newLeaves);
+    const newRoot = tree.get_root();
+    const formattedRoot = bufferToFixed(newRoot);
+
+    // compare root against contract, and store if there is a match
+    if (this._contract.isKnownRoot(formattedRoot)) {
```
drewstone, line 140: Great stuff mate

packages/contracts/src/contracts/anchor.ts
```diff
@@ -110,26 +107,46 @@ export class AnchorContract {
...
+      this.mixersInfo.setMixerStorage(this._contract.address, lastQueriedBlock, [
+        ...storedContractInfo.leaves,
+        ...fetchedLeaves.newLeaves,
+      ]);
+    }
+
+    let leafIndex = [...storedContractInfo.leaves, ...fetchedLeaves.newLeaves].findIndex(
+      (commitment) => commitment == bufferToFixed(deposit.commitment)
```
drewstone, line 148: @AhmedKorim @nepoche since the user is providing a deposit to withdraw, I would suggest we find a way for the user to also possess their deposit leaf's index a priori. This would allow us to remove this call entirely which may improve speed as we grow and at least simplify some things.

For example, as we work through this more, consider all the ways in which the user can provide information in place of us querying it. The protocol is guaranteed to work if the user provides the right inputs, therefore if they mess up or provide the wrong inputs it will simply fail.. the burden becomes more on the user.

drewstone (commented):
Looks great so far as well!

packages/contracts/src/contracts/anchor.ts
nepoche: Added this

packages/contracts/src/contracts/anchor.ts
nepoche: If it is a piece of data not completely controlled via the client, I think we could run into some issues with transaction ordering.

AhmedKorim (approved):
Looks good

PR #149, opened 2021-09-02
Implement cancelling on the withdraw tx

nepoche:


packages/contracts/src/contracts/anchor.ts
```diff
@@ -159,21 +175,20 @@ export class AnchorContract {
     return { proof, input: zkpInput };
   }
 
-  async withdraw(deposit: Deposit, zkpInputWithoutMerkleProof: ZKPInputWithoutMerkleProof) {
+  async withdraw(proof: any, zkp: ZKPInput) {
```
drewstone, line 178: Maybe we can enforce proof to be `Uint8Array`?

drewstone: Still looks okay, just thinking through as much validation as possible.

drewstone (commented):
Looks good so far, if we keep console logs inside app then we should opt to use the dedicated logger.

PR #165, opened 2021-09-17
Add proof generation from leaves response of active relayer

nepoche:
Move websocket connection initialization until it is needed.

Allow proof generation on leaves queried from the selected relayer to improve speeds.


drewstone (changes requested):
Looks great! We should add validation to the merkle root generated by the leaves from the relayer in a future PR.

PR #181, opened 2021-10-07
Support withdraw flow starting from any chain

nepoche:


packages/contracts/src/contracts/webb-anchor.ts
```diff
@@ -161,7 +180,7 @@ export class WebbAnchorContract {
   async generateMerkleProof(deposit: Deposit) {
     const bridgeStorageStorage = await bridgeCurrencyBridgeStorageFactory();
     const storedContractInfo: MixerStorage[0] = (await bridgeStorageStorage.get(this._contract.address)) || {
-      lastQueriedBlock: anchorsStorage[this._contract.address.toString()] || 0,
+      lastQueriedBlock: 0,
```
drewstone, line 183: Why not use from storage?

packages/contracts/src/contracts/webb-anchor.ts
```diff
@@ -196,56 +215,14 @@ export class WebbAnchorContract {
...
-    const tree = new MerkleTree('eth', treeHeight, [], new PoseidonHasher());
-    const lastQueriedBlock = storedContractInfo.lastQueriedBlock;
-
-    logger.trace('generate merkle proof with deposit', deposit, ` for the target roof ${targetRoot}`);
-
-    // Then, check the new leaves
-    const fetchedLeaves = await this.getDepositLeaves(lastQueriedBlock + 1, 0);
+  async generateLinkedMerkleProof(sourceDeposit: Deposit, sourceLeaves: string[]) {
+    // Get the source latest root
+    const latestSourceRoot = (await this._contract.getLatestNeighborRoots())[0];
```
drewstone, line 220: A more sustainable long term way to do this is to read the edges first. Then find the edge index of the source chain. Then grab the neighbor root at that index. Thoughts?

packages/apps/src/configs/relayer-config.ts
```diff
@@ -18,8 +18,10 @@ export function relayerNameToChainId(name: string): ChainId {
   switch (name) {
     case 'beresheet':
       return ChainId.EdgewareTestNet;
-    case 'harmony':
+    case 'harmonys1':
```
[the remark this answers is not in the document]
nepoche: Yes, we could - But we need to also do this on the relayer so we can match these values with the info response from the relayer. We may need to come up with some definitive naming scheme for the relayers.

PR #129, opened 2021-07-26
Webb relayer integration

AhmedKorim:

Issue #243, opened 2022-01-12
[TASK] Fix and remove logs / errors in browser console


drewstone:
## Overview 

When running the Dapp locally or visiting the development site, if you open dev tools in your preferred browser you will see a number of logs and react-specific errors polluting the console. 

We should go through each of these logs / errors and clean them up accordingly. Ideally we keep the browser console clean so that we can important errors / unexpected behaviour more clearly. 

**Below is a list of logs from opening the browser console on the development site:**

- [x] Update / fix `Warning: Invalid DOM property `stroke-width`. Did you mean `strokeWidth`?`
- [x] Update / fix `Warning: Invalid DOM property `stroke-linecap`. Did you mean `strokeLinecap`?`
- [x] Update / fix `Warning: Invalid DOM property `stroke-linejoin`. Did you mean `strokeLinejoin`?`
- [x] Remove console logs from react-environment/src/WebbProvider [here](https://github.com/webb-tools/webb-dapp/blob/4fee717a3ecd548aa5a6573b3da15f29e4c83841/packages/react-environment/src/WebbProvider.tsx#L81), [here](https://github.com/webb-tools/webb-dapp/blob/develop/packages/react-environment/src/WebbProvider.tsx#L332)
- [x] Fix `react-dom.development.jss?3c4a:88 Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components`
- [x] Fix `Warning: Can't perform a React state update on an unmounted component`
- [x] Update `Warning: findDOMNode is deprecated in StrictMode`
- [x] Fix `Warning: Each child in a list should have a unique "key" prop.`
- [x] Update `Warning: Failed prop type: Material-UI: `overlap="rectangle"` was deprecated. Use `overlap="rectangular"` instead.`
- [x] Update `Warning: Invalid DOM property `fill-rule`. Did you mean `fillRule`?`
- [x] Remove console logs from react-components [here](https://github.com/webb-tools/webb-dapp/blob/4fee717a3ecd548aa5a6573b3da15f29e4c83841/packages/react-components/src/Sidebar/Products.tsx#L88), and [here](https://github.com/webb-tools/webb-dapp/blob/4fee717a3ecd548aa5a6573b3da15f29e4c83841/packages/react-components/src/Sidebar/Products.tsx#L127)
[...]

PR #260, opened 2022-01-20
Relay substrate withdrawing transaction

AhmedKorim:
## PR Checklist
Please check if your PR fulfills the following requirements:
Should be merged after https://github.com/webb-tools/relayer/pull/86

- [ ] Tests for the changes have been added (for bug fixes/features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

Adding withdrawal submission throw relayers for substrate mixer

Issue Number: #129 

## What is the new behavior?
Users will be able to utilize relayers for withdrawing tx

- [ ] Yes
- [ ] No


packages/mixer/src/components/Withdraw/Withdraw.tsx
```diff
@@ -129,7 +131,12 @@ export const Withdraw: React.FC<WithdrawProps> = () => {
         }
       });
   };
-
+  const relayers = relayersState.relayers;
+  const activeRelayer = relayersState.activeRelayer;
+  console.log({
+    relayersState,
+    relayers,
+  });
```
nepoche, line 139: Can we do some logging cleanup / leave ones in that are helpful?

packages/react-environment/src/webb-context/relayer/webb-relayer.class.ts
```diff
@@ -116,9 +133,9 @@ export class WebbRelayerBuilder {
           : new Map(),
         substrate: info.substrate
           ? Object.keys(info.substrate)
-              .filter((key) => info.substrate[key]?.account && nameAdapter(key, 'evm') != null)
+              .filter((key) => info.substrate[key]?.beneficiary || nameAdapter(key, 'substrate') != null)
```
nepoche, line 136: Will the relayer for substrate always return `beneficiary`? Or is it still optional and if a beneficiary is not set in the config, will it return `account`?

packages/react-environment/src/webb-context/mixer/mixer-withdraw.ts
```diff
@@ -63,7 +63,7 @@ export abstract class MixerWithdraw<T> extends EventBus<MixerWithdrawEvents> {
   }
 
   // todo switch to the reactive api
-  get relayers(): Promise<WebbRelayer[]> {
+  get relayers(): Promise<WebbRelayer[]> | WebbRelayer[] {
```
nepoche, line 66: This seems strange, shouldn't it be one or the other?

nepoche (changes requested):

packages/react-environment/src/webb-context/relayer/webb-relayer.class.ts
AhmedKorim: I believe the `account` the field is deprecated in favor of the `beneficiary`, And relayer that supports substrate will have the `beneficiary` key  not account that's why I just added just `beneficiary` and it wont hurt adding `account` as well if you think so

packages/mixer/src/components/Withdraw/Withdraw.tsx
AhmedKorim: removed, will add a cleanup task for console logs from the whole dApp with #243

packages/react-environment/src/webb-context/relayer/webb-relayer.class.ts
nepoche: Nah it's better not to support it. Much better idea for it always to be beneficiary. Would you mind just adding a comment above line 127 that account is to support Mainnet Harmony and Shiden relayer deployments?

AhmedKorim: This is why I didn't add it @nepoche! , will restore to what I first implemented and add the comment you suggested

Issue #259, opened 2022-01-19
[SPEC] Bridge Config Modifications


drewstone:
### Overview

Other configuration values, such as currency-config and chain-config are accessible via the AppConfig and are quite modular. Whereas the bridge-config is quite complicated with nested structures.

We should break up the bridge-config into more modular parts - a bridge needs to only know about its BridgeCurrency, and the BridgeCurrency can have configured values:

```
// addresses of the wrapped token n each chain it lives on
const webbETHtest1 = new BridgeCurrency({
	[ChainId.Ropsten]: '0x2C165572FeBE99C25644eF3433989D6e37F2a8bE',
	[ChainId.Rinkeby]: '0x80DF11835B03bE444fd6f6587bEa70175d12da39',
	[ChainId.Goerli]: '0x021C3Ad0971c10f39a36E783419A491FB2dE3f64',
	[ChainId.Kovan]: '0xdd1CcAb2fB769A0D07bb9795F309964a12A79F7D',
	[ChainId.OptimismTestnet]: '0x33Ae40e485fcb8fD196f22229ba094F2b1680Bb7',
	[ChainId.ArbitrumTestnet]: '0x95A5CcfCa50F7Dcf410CE397Bb7Becf1b42Ae601',
},
"bridge-name"
);

// addresses of fixed anchor contracts defined over wrapped tokens
const AnchorConfig = {
	[webbETHtest1.name]: [{
		amount: '0.1',
	  anchorAddresses: {
		  [ChainId.Ropsten]: '0x8DB24d0Df8cc4CEbF275528f7725E560F50329bf',
		  [ChainId.Rinkeby]: '0x99285189A0DA76dce5D3Da6Cf71aD3f2b498DC88',
		  [ChainId.Goerli]: '0xC44A4EcAC4f23b6F92485Cb1c90dBEd75a987BC8',
		  [ChainId.Kovan]: '0xd961d7Cf4d001EC57ff3F6F9F6428B73b7d924Bc',
		  [ChainId.OptimismTestnet]: '0xd2e52699762D00f142e2c61280cd87D47B3A3b97',
		  [ChainId.ArbitrumTestnet]: '0x626FEc5Ffa7Bf1EE8CEd7daBdE545630473E3ABb',
		}],
	},
}
```

We can build up data on app load to satisfy the responsibility that a `BridgeConfig` has in our app, using something like the following:

```
export const bridgeConfigByChain: AppConfig['bridge-chains'] = {
[...]

PR #267, opened 2022-01-21
Refactor Bridge and BridgeCurrency Configuration

nepoche:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes/features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

Bridge currency and bridge configuration is treated differently than other configurations (chains, currencies, wallets). The Bridge configuration object is also complex and leads to complicated code in other areas of the codebase.

Issue Number: #259 

## What is the new behavior?
The bridge currency is now an entry in the `WebbCurrencyId`.  BridgeCurrencies will now be treated as a `Currency`.  A `type` property has been added to the `CurrencyView` to differentiate currencies in the `CurrencyConfig`.

Changes made in line with the spec.

Changes in other areas of the codebase required for the two above changes.

- [ ] Yes
- [x] No


packages/apps/src/configs/currencies/currency-config.ts
```diff
@@ -15,55 +18,122 @@ export const currenciesConfig: AppConfig['currencies'] = {
...
   },
   [WebbCurrencyId.ETH]: {
     name: 'Ethereum',
     symbol: 'ETH',
     color: '',
     id: WebbCurrencyId.ETH,
+    type: CurrencyType.WrappableEvmCurrency,
     icon: React.createElement(EtherLogo),
+    addresses: new Map([
+      [ChainId.Ropsten, '0x0000000000000000000000000000000000000000'],
```
drewstone, line 42: Let's use a `const ZERO_ADDRESS` so we can re-use a variable here instead.

packages/react-environment/src/types/currency-config.interface.ts
```diff
@@ -0,0 +1,21 @@
+import { ChainId, WebbCurrencyId } from '@webb-dapp/apps/configs';
+import { createElement } from 'react';
+
+export enum CurrencyType {
+  WrappableEvmCurrency,
```
drewstone, line 5: Can we document this and give examples? Recall that all non-native Substrate tokens currently are also `ORML` tokens so want to make sure semantically this makes sense.

drewstone: i.e. what are `WrappableEvmCurrency` examples, `ORML`, `BridgeCurrency`.

drewstone (changes requested):
LGTM pending some comments/updates.

packages/react-environment/src/types/currency-config.interface.ts
nepoche: Good idea, actually this should be updated to
```
enum CurrencyType {
     Erc20,
     ORML,
     BridgeErc20,
}
```

with a `type WrappableEvmCurrency = Erc20`.

The intent of the WrappableEvmCurrency is to easily distinguish which currencies can be selected in a 'wrap/unwrap' context with the other edge being a bridge currency.

nepoche: Good idea, actually this should be updated to
```
enum CurrencyType {
     Erc20,
     ORML,
     BridgeErc20,
}
```
with a type `WrappableEvmCurrency = Erc20`.

The intent of the WrappableEvmCurrency is to easily distinguish which currencies can be selected in a 'wrap/unwrap' context with the other edge being a bridge currency.

drewstone: Consider `ERC20, ORML, NATIVE`

packages/react-environment/src/types/chain-config.interface.ts
```diff
@@ -11,5 +11,5 @@ export interface ChainConfig {
   blockExplorerStub?: string;
   logo: React.ComponentType;
   nativeCurrencyId: WebbCurrencyId;
-  currencies: Array<{ currencyId: WebbCurrencyId; enabled: boolean; address: string }>;
```
AhmedKorim, line 14: I like cleaning this up but to be just `WebbCurrenchID` however, it seems the enabled key could be useful to disable the asset for some reason, WTDYT?

AhmedKorim, line 14: I like cleaning this up but to be just `WebbCurrenchID` however, it seems the enabled key could be useful to disable the asset for some reason, WTDYT?

packages/react-environment/src/webb-context/bridge/bridge.ts
```diff
@@ -1,75 +1,51 @@
...
-    const tokens = Bridge.getTokens(config);
-    return tokens.filter((token) => {
-      for (let chainId of chainIds) {
-        if (!token.hasChain(chainId)) {
-          return false;
-        }
-      }
-      return true;
-    });
-  }
```
AhmedKorim, line 70: @nepoche  I see this can be a useful utility to get tokens that are among more than one chain, but I support removing it if we don't rely on static configuation to tell if the asset is Bridgeable

nepoche: This method wasn't being used anywhere so I removed it. We can always add back a similar method if it is useful in the future.

packages/react-environment/src/types/chain-config.interface.ts
nepoche: I don't know what is the use case for disabling an asset (only in UI) for the time being. We can add it back later if it is useful.

Issue #395, opened 2022-05-19
[TASK] Implement Deposit and Withdraw TX processing page UI


dutterbutter:
## Overview 

Currently when users select 'Deposit' on the bridge page, they are presented with their secret note that they must confirm before selecting 'Deposit' again to actually deposit the funds into the bridge. However, with the VAnchor we should be able to provide an improved experience by eliminating the need to select Deposit twice. 

Upon selecting Deposit, we should present the user with a TX processing modal similar to the below mock design. In addition, you may view a similar modal on Tornado Nova [here](https://docs.tornado.cash/tornado-cash-nova/fund-and-withdraw-on-nova). 

![Single-Chain_Deposit_Modal_VAnchor_v0 0 1](https://user-images.githubusercontent.com/29983536/169304792-ad91056b-727b-4d7e-9d9b-30b4497b1338.png)

### UI Elements

- [ ]  A modal pops up upon selecting Deposit
- [ ] Background behind modal is blurred, and is the focus of the view
- [ ] A sequential set of steps should be listed reflecting the state of the tx process (e.g. Generating Proof, Sending Transaction etc)
- [ ] Once transaction is successful it should indicate the Deposit has been made successfully with meta data (e.g. deposited amount)

dutterbutter:
Closed in #442

PR #442, opened 2022-06-22
Add the base for a transaction processing modal over VBridge components

nepoche:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

Upon confirming a deposit, the only indication given of its progress is a pending notification on the side of the app.

Issue Number: #395 

## What is the new behavior?
Implementation of a `TransactionProcessingModal` which provides more detailed information to the user in VAnchor flows (since deposits require ZKP generation, as well as withdraws - while deposits in FixedAnchor or Mixer do not require ZKP gen).

- [ ] Yes
- [x] No


drewstone:
Looks great!

drewstone (approved):

dutterbutter:
A couple things I noticed that would be good to resolve either in this PR or subsequently:

- [ ] If a user makes a deposit, and selects 'Hide' on the modal there is no way to unhide the modal 
- [ ] If a user makes a deposit, and selects 'Hide' on the modal, and then proceeds to make a second deposit the modal remains hidden (e.g. state remains closed) 
- [ ] We should remove the right side notification in favour of this modal 

Feel free to comment which issues above will be handled in this PR or a subsequent one.

PR #496, opened 2022-08-15
ECDSA Claims page

### LINKED ISSUE #480: [TASK] ECDSA Claims Interface
## Overview

We have integrated a `Claims` pallet that allows us to set up a genesis distribution for Ethereum address holders. The basic mechanism is that a valid Ethereum address would generate a signature of their Substrate address to receive the distribution. We run the following check on-chain:

1. Is the Ethereum address in the set of recipients
2. Is the signature valid for that address
3. If so, send vested/unvested tokens to the target Substrate address

## Task Checklist

- [x]  Create a UI that takes as input a Substrate address raw 32-bytes public key (to be checked against the pallet)
- [x]  The first step in the process should be to sign the Substrate address through MetaMask with your valid Ethereum key
- [x]  The second step should be to submit this signature and tx to the Substrate network claims pallet.
- [ ]  The third step should be to show confirmation and a tx hash.

AhmedKorim:
## Summary of changes
- Getting the claim call working on the dApp
- Added the ablity to send unsigned tx for Polkadot 

- Closes #480 

### Demo
https://user-images.githubusercontent.com/35463181/185615174-cbd10fe0-c967-4ff7-9a6f-1de300fd4d43.mp4

-----
### Code Checklist 

- [x] Tested

- [ ] Documented


packages/api-providers/src/ext-providers/web3/web3-provider.ts
```diff
@@ -197,4 +203,11 @@ export class Web3Provider<T = unknown> {
       },
     });
   }
+
+  async sign(payload: `0x${string}`, account: string): Promise<string> {
```
nepoche, line 207: Can we move the logic of this function to the WebbWeb3Provider's `sign` function, and rename that function to `signSubstrateAddressForClaim`? The way it is now, it seems like this is supposed to be a generic signing function when it is actually very specific.

AhmedKorim: Sure sounds good

AhmedKorim: Made the `sign` function generic now just takes a string, the substrate claims logic now in the hooks

nepoche (approved):

MohamedElmdary:
Great job  as always @AhmedKorim <3

Issue #413, opened 2022-06-02
[SPEC] Optimize leaf fetching in `ApiProvider`


dutterbutter:
## Overview

The current flow `ApiProvider` function consists of:

1. Group utxos by insertion chain
2. Fetch leaves for all insertion chains
3. Create `VAnchorPMSetupInput` and send to `sdk-core` for a proof

The fetching of leaves operation should be reviewed and a new optimized approach should be considered. One potential solution discussed was add multi-threaded support. 

- [ ] Outline potential optimized solutions
- [ ] Benchmark existing approach vs new approach

dutterbutter:
@nepoche please provide any additional details to this task as you see fit. Thanks.

nepoche:
Currently for proof generation with multiple different source chains, leaves are fetched sequentially instead of at the same time. We should fetch leaves from all source chains at the same time as an optimization.

PR #498, opened 2022-08-18
Optimize leaf fetching

AtelyPham:
- ci: bump `webb.js` to `0.1.4-95`
- perf: optimize leaes fetching in parallel
- perf: optimize leaves fetching in parallel

## Summary of changes
- Bumped `webb.js` to `0.1.4-95`
- Extract the leaf fetching into the particular method to support fetching leaves for multiple notes simultaneously instead of sequentially.

- Closes https://github.com/webb-tools/webb-dapp/issues/413

### Some benchmarks

- 2 notes:

CleanShot 2022-08-18 at 17 16 24@2x

- 5 notes:

CleanShot 2022-08-18 at 17 25 17@2x

- 7 notes:

CleanShot 2022-08-18 at 17 26 43@2x

-----
### Code Checklist 

- [x] Tested
- [ ] Documented


packages/api-providers/src/web3/vanchor-withdraw.ts
```diff
@@ -450,4 +401,68 @@ export class Web3VAnchorWithdraw extends VAnchorWithdraw<WebbWeb3Provider> {
...
+      const sourceHttpProvider = Web3Provider.fromUri(sourceChainConfig.url);
+      const sourceEthers = sourceHttpProvider.intoEthersProvider();
+      const sourceVAnchor = this.inner.getVariableAnchorByAddressAndProvider(sourceAddress, sourceEthers);
+      const leafStorage = await bridgeStorageFactory(Number(parsedNote.sourceChainId));
+      const leaves = await this.inner.getVariableAnchorLeaves(sourceVAnchor, leafStorage, abortSignal);
+
+      leavesMap[parsedNote.sourceChainId] = leaves.map((leaf) => {
+        return hexToU8a(leaf);
+      });
+    }
```
AhmedKorim, line 429: The mutation of the `leavesMap` object is being done with many async calls this can make undermind behavior, and some values could be overwritten.
it will be better to prepare the map with default values before mutation, and while mutating consider adding to the value
 **EX :**
This call 
```typescript
leavesMap[parsedNote.sourceChainId] = leaves.map((leaf) => {
        return hexToU8a(leaf);
      });
```
could be 
```typescript
leavesMap[parsedNote.sourceChainId] =[ ...leavesMap[parsedNote.sourceChainId],leaves.map((leaf) => {
        return hexToU8a(leaf);
      })];
```

Besides the fact that you may fetch the same leaves twice

AhmedKorim: @nepoche  wanna know you thoughts as well

AhmedKorim (changes requested):

packages/api-providers/src/web3/vanchor-withdraw.ts
nepoche: I agree that the check:
```
    // fetch leaves if we don't have them
    if (leavesMap[parsedNote.sourceChainId] === undefined) {
```
will not prevent multiple calls for fetching leaves of the same chain, and the leavesMap[parsedNote.sourceChainId] will be overwritten with leaf data multiple times.  However, the data should always be all leaves 1x.  With the 2nd option you listed, the data would be all the leaves potentially 1-15x.

drewstone: I think the proposal from @AhmedKorim will lead to more bugs, it could end up appending lots of duplicate leaves. I do think that there's room for improvement on ensuring we only fetch leaves for each unique chain once.

We could go through all notes and get the unique set of chains they're originating/targeting **and then** begin querying each one. This would ensure we don't double-start queries from the same chain before either finishes.

drewstone: Can we try and put each chain query into a WebWorker? Do you think that would help parallelize and not block the main thread @AtelyPham ?

drewstone (approved):
I think this is good improvement nonetheless, I support merging. We can create another improvement task to try and make it faster using another route. CC @dutterbutter

packages/api-providers/src/web3/vanchor-withdraw.ts
AtelyPham: Yeah, I agree that we can prepare a set of unique source chains and then fetch leaves from the prepared set. About the `WebWorker`, I think we don't really need the WebWorker, because the main thread still needs the result of the parallel computations to continue.

AhmedKorim (approved):

Issue #509, opened 2022-08-19
[TASK] Implement reusable Chip component


dutterbutter:
## Task Description

The purpose of this task is to implement a reusable Chip component that can be exported for external use. Chips are compact elements that represent a state, attribute, or action.

## Task Deliverable

- [x]  Create a folder `Chip`
- [x]  Implement a Chip component with the below mentioned props
- [x]  Ensure the implementation reflects the intended design
- [x]  Document each component with an example
- [x]  Ensure component is exportable

### Chip Component

The are variant types of `Chip` that are dictated by the props passed in. Below is a list of the components properties that are required. Please refer to the `Chip` design system link at the bottom to view each different chip type that are made up in this component. 

**Component properties:**

- color="green | blue | purple | yellow | red";
- state="default | disabled";
- dark mode="off | on";

**Example usage:**

```bash
<Chip
  color="green"
  dark-mode="off"
  state="default"
/>
```

## Relevant links

- [Figma Design System](https://www.figma.com/file/H43lYnTCOAPjO9kTrrepe1/%F0%9F%95%B8%EF%B8%8F-Design-System?node-id=164%3A4609)

PR #544, opened 2022-09-05
New Chip component

AtelyPham:
## Summary of changes
- Add a new `Chip` component

- Closes https://github.com/webb-tools/webb-dapp/issues/509

-----
### Code Checklist 

- [x] Tested
- [x] Documented

### Screenshots

CleanShot 2022-09-05 at 15 47 46@2x
CleanShot 2022-09-05 at 15 47 55@2x

### DIFF
--- a/packages/webb-ui-components/src/components/Chip/Chip.tsx
+++ b/packages/webb-ui-components/src/components/Chip/Chip.tsx
@@ -1,0 +1,18 @@
+import React, { useMemo } from 'react';
+import { twMerge } from 'tailwind-merge';
+
+import { ChipProps } from './types';
+import { getChipClassName } from './utils';
+
+export const Chip: React.FC<ChipProps> = (props) => {
+  const { children, className, color = 'green', isDisabled } = props;
+
+  const baseClsx = useMemo(() => 'box-border inline-block px-3 py-1 rounded-full font-bold body4', []);
+
+  const _className = useMemo(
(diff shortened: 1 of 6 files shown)


packages/webb-ui-components/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,33 @@
...
+ *  <Chip>Active</Chip>
+ *  <Chip color="red" isDisabled>Disabled</Chip>
+ * ```
+ */
+export const Chip: React.FC<ChipProps> = (props) => {
+  const { children, className, color = 'green', isDisabled } = props;
+
+  const baseClsx = useMemo(() => 'box-border inline-block px-3 py-1 rounded-full font-bold body4', []);
+
+  const _className = useMemo(
```
AhmedKorim, line 27: `_` an underscore prefix for variable means it isn't being used, can we change this to be something else or change the destructuring variable name for `className` to be `class` or smt

AhmedKorim (changes requested):

packages/webb-ui-components/src/components/Chip/Chip.tsx
AhmedKorim: @AtelyPham  other than that looks good

AhmedKorim (approved):

PR #556, opened 2022-09-09
Fix package updates and tx-check

AhmedKorim:
## Summary of changes
- 

- Closes 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented


packages/api-providers/src/polkadot/vanchor-deposit.ts
```diff
@@ -205,6 +205,8 @@ export class PolkadotVAnchorDeposit extends VAnchorDeposit<WebbPolkadot, Deposit
         inputNotes: [inputNote],
         publicAmount,
         output: [output1, output2],
+        refund: '0',
```
nepoche, line 208: Add to the extData above as well.

PR #585, opened 2022-09-20
SubQuery timetamp

### LINKED ISSUE #577: [SPEC] Stats using webb-subql time
## Overview
The SubQuery node may suffer from delays from the node we should depend on its timing instead of using the `Date.now()`

### 1. Pulling the latest synced block with the block height interval
```graphql
query LastBlock{
  blocks(first: 1, orderBy:[NUMBER_DESC]) {
    nodes {
      timestamp
      number
    }
  }
}

```

Create a react provider that will share the SubQuery node status and a custom time object

**Provider config**
```typescript
	type ProviderConfig = {
		// Number of seconds for a block to be generated
		blockTime:number;
		// Number of blocks for a session
		sessionHeight:number
	}
```
```typescript 
/**
 * Wrapper date object that consumes the SubQuery node blocks fetching as a time source
 * */
class SubQlTime {
  private lastSynced: Date;

  constructor(private lastBlockTime: Date) {
    this.lastSynced = new Date();
  }
  /**
   * Sync the inner date object
   * */
  sync(date: Date) {
    this.lastBlockTime = date;
    this.lastSynced = new Date();
  }

  get current(): Date {
    // Calculate the different from the sync point
    const diff = new Date().getTime() - this.lastSynced.getTime();
    // add the diff to the block timestamp
    return new Date(this.lastBlockTime.getTime() + diff);
  }

  syncFromIsString(dateStr: string) {
    const date = new Date(dateStr);
    this.sync(date);
  }
}

```

AhmedKorim:
## Summary of changes
-  Adding a provider to share the current block number and other metadata
- Creating a custom time object that gets synchronized from SubQuery's last fetched block 

- Closes #577 
image

-----
### Code Checklist 

- [x] Tested
- [x] Documented

### DIFF
--- a/packages/page-statistics/src/App.tsx
+++ b/packages/page-statistics/src/App.tsx
@@ -1,4 +1,5 @@
-import { ApolloClient, InMemoryCache } from '@apollo/client';
+import { ApolloClient, ApolloProvider, InMemoryCache } from '@apollo/client';
+import { StatsProvider } from '@webb-dapp/page-statistics/provider/stats-provider';
 import { RouterProvider } from '@webb-dapp/react-environment/RouterProvider';
 import { WebbUIProvider } from '@webb-dapp/webb-ui-components/provider';
 import { FC } from 'react';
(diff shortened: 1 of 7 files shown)


nepoche (approved):

AtelyPham (approved):
LGTM 👍

PR #601, opened 2022-09-21
Include the previous and next  entity id for single view

### LINKED ISSUE #600: [TASK] Return next and previous for single views
# Overview
The single view (Details page) has the next and previous
image

The goal for single view hooks to return if there's a next/previous and the id for each item

AhmedKorim:
## Summary of changes
- For Proposals and keys there is a new `next` `prev` ids to fetch the next item if any
- The `next` `prev` can be null which means there is no previous or next entry

- Closes #600 

-----
### Code Checklist 

- [x] Tested
- [x] Documented


AtelyPham (approved):
LGTM 👍

PR #610, opened 2022-09-23
Fix session and keys data

### LINKED ISSUE #609: [TASK] Stats keys page serving up incorrect data
### Description of issue
Reviewing the latest keys page the data being served up is very incorrect. 

Using the image below, the following areas need to be fixed:
- The session should be the latest session which is not 10 - this looks to be a block number which is incorrect 
- The key being used is uncompressed when we should be using the compressed key
- The progress bar are not working 
- Date is wrong 
- Details link never resolves
Screen Shot 2022-09-22 at 5 13 35 PM

Using the image below, the following areas need to be fixed:
- The current keys implementation table displays keys in random intervals, this should default to from latest
- The authority set is blank 
- The key depicted should be the compressed key
- The thresholds are not displayed correctly 
- The sessions are not accurate it looks to be using a block number instead 
screenshot

AhmedKorim:
## Summary of changes
- Update the query for last session
- Update the stats provider to use session height to 600

**Note** this requires the merging/deploying of the subquery node https://github.com/webb-tools/webb-subql/pull/23
- Closes #609 
-----
### Code Checklist 

- [ ] Tested
- [ ] Documented


dutterbutter:
@AhmedKorim this currently does not close #609

AhmedKorim:
https://user-images.githubusercontent.com/35463181/192275883-b2c9c476-7751-4e8a-83fd-58a8a5f769b1.mp4

@dutterbutter

packages/page-statistics/src/containers/KeygenTable/KeygenTable.tsx
```diff
@@ -140,10 +147,9 @@ export const KeygenTable: FC = () => {
     [pageSize, pagination.pageIndex, pagination.pageSize]
   );
 
-  const pageCount = useMemo(() => Math.floor(totalItems / pageSize), [pageSize, totalItems]);
+  const pageCount = useMemo(() => Math.round(totalItems / pageSize), [pageSize, totalItems]);
```
nepoche, line 150:
```suggestion
  const pageCount = useMemo(() => Math.ceil(totalItems / pageSize), [pageSize, totalItems]);
```

packages/page-statistics/src/provider/stats-provider.tsx
```diff
@@ -20,22 +20,20 @@ export type Metadata = {
...
  *
  * Get the next session from metadata
  * */
-export function nextSession(height: string): string {
+export function nextSession(height: string, sessionHeight: number): string {
   const blockNumber = Number(height);
-  const sessionNumber = Math.floor(blockNumber / 10) * 10;
-
-  return String(sessionNumber);
+  const sessionNumber = Math.round(blockNumber / sessionHeight) + 1;
```
nepoche, line 35:
```suggestion
  const sessionNumber = Math.ceil(blockNumber / sessionHeight);
```

packages/page-statistics/src/containers/KeygenTable/KeygenTable.tsx
nepoche, line 150:
```suggestion
  const pageCount = useMemo(() => Math.ceil(totalItems / pageSize), [pageSize, totalItems]);
```

nepoche (changes requested):

packages/page-statistics/src/containers/KeygenTable/KeygenTable.tsx
AhmedKorim: You are right

AtelyPham: This happened because last time we didn't handle the `null` case so some latest sessions may contain `null` values. But at the moment if we have dealt with the `null` case, we should update the `pageCount` calculation.

packages/webb-ui-components/src/components/Avatar/Avatar.tsx
```diff
@@ -32,6 +32,7 @@ export const Avatar: React.FC<AvatarProps> = (props) => {
     darkMode,
     fallback,
     size = 'md',
+    sourceVariant,
```
AtelyPham, line 35: `sourceVariant` = 'uri'

AhmedKorim: It's an optional prop, no need for a default value

AtelyPham (approved):
LGTM 👍

packages/webb-ui-components/src/components/Avatar/Avatar.tsx
```diff
@@ -32,6 +32,7 @@ export const Avatar: React.FC<AvatarProps> = (props) => {
     darkMode,
     fallback,
     size = 'md',
+    sourceVariant = 'uri',
```
AtelyPham, line 35: Sorry! But do you think we need to change the default `sourceVariant` is `address`? Because right now, most of our implementation is using address. @AhmedKorim

AhmedKorim: it's okay to change them the occurrence did it on other PRs

PR #612, opened 2022-09-26
Remove global styles from dapp and api connection error handling

nepoche:
## Summary of changes
- 

- Closes 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented


packages/api-providers/src/abstracts/currency/orml-currency.ts
```diff
@@ -18,7 +18,14 @@ export class ORMLCurrency {
...
   async list() {
-    const assets = await this.api.api.query.assetRegistry.assets.entries();
+    let assets;
+
+    try {
+      assets = await this.api.api.query.assetRegistry.assets.entries();
+    } catch (ex) {
+      console.log('tried to get assets from a chain without support');
+      return [];
+    }
```
AhmedKorim, line 28: If you can import types and get the right values instead of `toHuman()` this would be better

And  you can check for the asset registry existence

```
const assetRegistry = this.api.api.query. assetRegistry;
if(!assetRegistry) {
console.log('tried to get assets from a chain without support');
return []
}
...
```

AhmedKorim (changes requested):

AhmedKorim:
@nathan looks good just that Polkadot call if we can use the types

packages/api-providers/src/abstracts/currency/orml-currency.ts
nepoche: I'm confused as to where the 'toHuman()` call you are referring to is in my PR.

AtelyPham (approved):
LGTM 👍

PR #673, opened 2022-11-01
chore: update github ci and readme

dutterbutter:
## Summary of changes
- Updates GH actions to only deploy previews for non draft pull requests to reduce resource exhaustion
- Removes on push deployment via GH actions as this is configured out of the box via the Netlify UI 
- Updates readme to reflect latest changes 

### Discussion

The readme does not include an overview of all the directories in `/lib`. There are currently 20 directories there and I think that might be excessive to include in a README where visitors just want to start the varying UI's. Instead I propose we create READMEs for each directory in `/libs` which provide context to its purpose and usage. 

@nepoche @AtelyPham @AhmedKorim please let me know your thoughts on the above. 
  
- Closes 

-----
### Code Checklist 

- [x] Tested
- [x] Documented


AtelyPham:
That sounds good to me. I think we can create a summary for all repos in the root README file with the permanent links to the README file of each repo.

AtelyPham (approved):
LGTM 👍

PR #714, opened 2022-11-22
Add wrap and deposit flow

### LINKED ISSUE #705: [TASK] Add wrap and deposit flow
## Overview 

Currently the deposit flow does not contain the wrap and deposit flow. A user should be able to select an asset (e.g. `ETH`) from the token selection, and then the dApp should detect that selection and move to a wrap and deposit form (see image below). If the user selects a bridged asset (e.g. `webbETH`) to deposit into the bridge, then we do not need to show the wrap and deposit form as that asset can be deposited directly into the bridge. The figma link below illustrates how the flow should look. 

Screen Shot 2022-11-21 at 4 04 16 PM

### Task Checklist 

- [x] Correctly detect non-wrapped asset 
- [x] Display wrap UI form showing the bridged asset
- [ ] Test wrap and deposit flow for each chain
- [ ] Test deposit flow for bridged asset

AhmedKorim:
## Summary of changes
- 

- Closes #705 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented

### DIFF
--- a/apps/bridge-dapp/src/containers/DepositContainer/DepositContainer.tsx
+++ b/apps/bridge-dapp/src/containers/DepositContainer/DepositContainer.tsx
@@ -25,6 +25,7 @@ import { ChainSelectionWrapper, WalletModal } from '../../components';
 import { CreateAccountModal } from '../CreateAccountModal';
 import { DepositConfirmContainer } from './DepositConfirmContainer';
 import { DepositContainerProps } from './types';
+import { DepositCardProps } from '@webb-tools/webb-ui-components/containers/DepositCard/types';
 
 export const DepositContainer = forwardRef<
   HTMLDivElement,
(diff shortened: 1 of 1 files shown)


dutterbutter (commented):
Nit: can you update the dir from: 

`libs/abstract-api-provider/src/anchor/bridge-api.ts` to this
`libs/abstract-api-provider/src/vanchor/bridge-api.ts`

AtelyPham (changes requested):
CleanShot 2022-11-23 at 18 41 57@2x

The bridging token does not seem correct, I think it should be `ETH/WEBBETH`. I think you should update the asset type [here](https://github.com/webb-tools/webb-dapp/blob/0957f9f4d3154e5e659258b983f87f6faa216717/libs/webb-ui-components/src/components/BridgeInputs/ShieldedAssetInput.tsx#L70-L71). @AhmedKorim

AhmedKorim:
it should be just WEBBETH only without the / 
so it's just one token

AhmedKorim:
image
Used a normal `TokenInput` instead of `ShieldedAssetInput`

AtelyPham (approved):
LGTM 👍

Issue #712, opened 2022-11-21
[CHECKLIST] Implement successful withdraw flow


dutterbutter:
## Overview 

The goal of this checklist is to create a comprehensive list of tasks that collectively make up the successful and unsuccessful withdraw flow for dApp users. 

### Withdraw Flow

**Note:** Assumes a user has already connected their wallet, created their note account, and made a deposit.

**Withdraw Scene-1 | Data Input** 
1. Selects token to withdraw from bridge
2. Select amount to be withdrawn 
3. Select relayer from relayer selection modal 
4. Input or paste recipient address 

There are additional flows to be considered for Withdraw. User can select 'Unwrap' using a toggle. The toggle should be disabled if there is no liquidity of the underlying asset to unwrap into, and enabled if there is liquidity to unwrap into. We will need to make a liquidity check during this operation. 

The other flow is the user can also, although not recommended entirely, to enter a variable amount to be withdrawn rather than a fixed amount. 

<img width="377" alt="Screen Shot 2022-11-21 at 4 46 21 PM" src="https://user-images.githubusercontent.com/29983536/203173341-a76920fb-056d-489c-b069-f9bf9ec6a1f8.png">

**Withdraw Scene-2 | Confirmation**
1. User confirms they have copied the new spend note 
2. User selects 'Withdraw' 

There is a case, where users do not have a new spend note generated, as they have withdraw the entire balance (e.g. spent all their notes). We need to account for this case here. 

<img width="377" alt="Screen Shot 2022-11-21 at 4 48 18 PM" src="https://user-images.githubusercontent.com/29983536/203173535-b7c2613c-a126-4165-bc8b-f6b04c1a4bd8.png">

**Withdraw Scene-3 | Withdraw In-progress**
1. Display the progress of the withdraw operations 
2. User can select 'New Transaction' that will move progress view into accordion component to the right side and display withdraw scene-1 

<img width="384" alt="Screen Shot 2022-11-21 at 4 50 06 PM" src="https://user-images.githubusercontent.com/29983536/203173838-7e1f7521-d5eb-467b-a788-9469c9b441e3.png">

**Right-sided component** - It should read Withdraw in the chip component not deposit.
<img width="283" alt="Screen Shot 2022-11-21 at 4 50 28 PM" src="https://user-images.githubusercontent.com/29983536/203173856-6b531a14-8385-4a20-adda-3b16d6498c86.png">

### Task Checklist 

[...]

PR #722, opened 2022-11-28
The Withdraw Flow

AtelyPham:
- correct the assset urls in note account table
- feat: update missing dependencies on deposit container
- chore: update withdraw flow state
- add error message for amount validation
- re-style fixed amount items
- add strict to tsconfig for react-hooks package
- add more states for withdraw confirm container
- re-style withdraw confirm component
- correct the relayer in the withdraw confirm card
- fix: the amount calculation
- style: update new design for the note section
- update the disabled logic inside withdraw confirm card
- add missing state inside remove all note function
- set gas limit manually, remove redundant await and cached new leaves
- allow withdraw and unwrap inside unwrap function
- update currency type in withdraw confirm container
- add transaction queue to withdraw flow
- remove notifications
- improve the note account creation animation
- fix: modify gas limit to fix the out of gas transaction error

## Summary of changes
- Add the explorer URL for assets in the note account table
- Re-style components
- Add error message for amount input
- Cached new fetched leaves to improve the leaves fetching function
- Add unwrap functionality into the withdraw function to allows withdraw and unwrap

- Closes https://github.com/webb-tools/webb-dapp/issues/712

-----
### Code Checklist 

- [x] Tested
- [x] Documented

### Additional info

- Right now, I only can deposit and withdraw on the same chain.
- Currently, we don't support withdraw and unwrap that works with relayers
- We also need to improve the transaction state for the `TransactionQueueCard` component to support multiple transactions at the same time.

### Screen Recording

https://user-images.githubusercontent.com/60747384/204351524-dbbec4a8-10dc-4a66-8769-fc53c1d048ee.mp4


libs/web3-api-provider/src/webb-provider/vanchor-withdraw.ts
```diff
@@ -420,16 +386,39 @@ export class Web3VAnchorWithdraw extends VAnchorWithdraw<WebbWeb3Provider> {
...
+              ...publicInputs,
+              outputCommitments: [
+                publicInputs.outputCommitments[0],
+                publicInputs.outputCommitments[1],
+              ],
+            },
+            extData,
+            unwrapTokenAddress,
+            {
+              gasLimit: 10000000 * 9,
```
AhmedKorim, line 403: @dutterbutter  can we have a task for improving gas limit calculation?

drewstone: Gas limit is fixed on these chains so really no need for calculation. Just need to keep a map and update it from time to time (these things update infrequently)

drewstone: Gas price is what changes and could be calculated

dutterbutter:
@AtelyPham please resolve conflicts

libs/web3-api-provider/src/webb-provider.ts
```diff
@@ -266,6 +265,12 @@ export class WebbWeb3Provider
       );
 
       leaves = [...storedContractInfo.leaves, ...leavesFromChain.newLeaves];
+
+      // Cached the new leaves
+      await storage.set(contract.inner.address.toLowerCase(), {
+        lastQueriedBlock: leavesFromChain.lastQueriedBlock,
+        leaves,
+      });
```
AhmedKorim, line 273: We need to validate these leaves to be right by building the tree and checking the root to be right, this is something we can iterate in a nother PR

apps/bridge-dapp/src/hooks/useTransactionProgressValue.ts
```diff
@@ -0,0 +1,57 @@
...
+        setProgress(null);
+        break;
+      }
+
+      default: {
+        throw new Error(
+          'Unknown transaction state in DepositConfirmContainer component'
+        );
+      }
+    }
```
AhmedKorim, line 53: Should be same as Ideal  or removed as we are matching against an enum

AhmedKorim (changes requested):

AhmedKorim (approved):

PR #771, opened 2022-12-05
Chips highlighting finalized best and session

### LINKED ISSUE #770: Add `Chips` highlighting Finalized, Best and Session

devpavan04:
## Summary of changes
- Adds chips highlighting finalized block, best block and session.

https://user-images.githubusercontent.com/53374218/205591422-e847caa9-e0a7-4cdb-9b2d-1c1aff867b11.mov

- Closes #770


apps/stats-dapp/src/provider/hooks/useBlocks.ts
```diff
@@ -0,0 +1,71 @@
...
+      isLoading: true,
+      isFailed: true,
+      error: undefined,
+    };
+  }, [bestBlock, finalizedBlock]);
+
+  useEffect(() => {
+    subscribeToNewHeads();
+    subscribeToFinalizedHeads();
+  });
```
AhmedKorim, line 52: The `useEffect` hook will be called every time the hook gets rendered if you don't pass the dependency array
Please write it like this
```
  useEffect(() => {
    subscribeToNewHeads();
    subscribeToFinalizedHeads();
  },[]);
```
This is going to call the `subscribeToNewHeads` and `subscribeToFinalizedHeads` once for every render if this hook, however, the two functions are doing subscription so a cleanup is required

you can write this hook as


```
  
  useEffect(() => {
    // Array of subscriptions
    const subscriptions: Array<() => void> = [];
    // The handier for listenting for new/best header
    const handler = async (): Promise<Array<() => void>> => {
      const unsubscribeNewHeads = await api.rpc.chain.subscribeNewHeads(
        (lastHeader) => {
          setBestBlock(lastHeader.number);
        }
      );
      const unsubscribeFinalizedHeads =
        await api.rpc.chain.subscribeFinalizedHeads((finalizedHeader) => {
          setFinalizedBlock(finalizedHeader.number);
        });
      return  [
        unsubscribeNewHeads,
        unsubscribeFinalizedHeads
      ]
    };

    handler().then((cleanup) => {
      subscriptions.push(...cleanup);
    });

    return () => {
      subscriptions.forEach((unsubscribe) => unsubscribe());
    };
  }, [api,setBestBlock,setFinalizedBlock]);
```

apps/stats-dapp/src/provider/stats-provider.tsx
```diff
@@ -148,10 +158,21 @@ export const StatsProvider: React.FC<Omit<StatsProvidervalue, 'isReady' | 'metaD
...
+  useEffect(() => {
+    getPromiseApi();
+  }, [polkadotProviderEndpoint]);
+
+  const getPromiseApi = async () => {
+    const wsProvider = new WsProvider(polkadotProviderEndpoint);
+    const apiPromise = await ApiPromise.create({ provider: wsProvider });
+    setApi(apiPromise);
+  };
+
```
AtelyPham, line 175: Consider moving this function inside the `useEffect` hook.

apps/stats-dapp/src/provider/stats-provider.tsx
```diff
@@ -148,10 +158,21 @@ export const StatsProvider: React.FC<Omit<StatsProvidervalue, 'isReady' | 'metaD
...
       metaData,
+      api,
     };
-  }, [staticConfig, metaData, isReady, time]);
+  }, [staticConfig, metaData, isReady, time, api]);
   const query = useLastBlockQuery();
 
+  useEffect(() => {
+    getPromiseApi();
+  }, [polkadotProviderEndpoint]);
```
AtelyPham, line 168: When you move the `getPromiseApi` into this `useEffect` hook, you don't need the `polkadotProviderEndpoint` in the dependency array because it has a static value.

apps/stats-dapp/src/constants/index.ts
```diff
@@ -3,3 +3,5 @@ const isDev = process.env.NODE_ENV === 'development';
 export const defaultEndpoint = isDev
   ? 'https://subquery-dev.webb.tools/graphql'
   : 'https://subquery-dev.webb.tools/graphql';
+
+export const polkadotProviderEndpoint = 'wss://arana-alpha-1.webb.tools';
```
AhmedKorim, line 7: @dutterbutter @devpavan04  this fine, but we should query the specName or the chain from the subquery node and make sure we are using the same endpoint as the subquery node so we don't end with different sources

```graphql
query{
  _metadata{
    chain
    specName
    
  }
}
```

```json
{
  "data": {
    "_metadata": {
      "chain": "Arana Alpha",
      "specName": "tangle-standalone"
    }
  }
}
```

apps/stats-dapp/src/provider/hooks/useBlocks.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+  };
+
+  const subscribeToFinalizedHeads = async () => {
+    if (api) {
+      await api.rpc.chain.subscribeFinalizedHeads((lastHeader) => {
+        setFinalizedBlock(lastHeader.number);
+      });
+    }
+  };
```
AtelyPham, line 68: Move these functions inside the `useEffect` hook (like the suggestion) to have the `unsubscribe` functions to be called when the component is unmounted.

AtelyPham (changes requested):
Some change suggestions @devpavan04. Let me know if you have any questions. Nice works 💯

apps/stats-dapp/src/provider/hooks/useBlocks.ts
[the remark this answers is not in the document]
AhmedKorim: use effect callback shouldn't be async and should add an empty array for deps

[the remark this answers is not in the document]
AtelyPham: My bad, your suggestion looks good so I'll resolve this comment.

apps/stats-dapp/src/constants/index.ts
dutterbutter: Yes let's avoid relying on hardcoded values as the stats dApp frontend should be usable for multiple webb networks.

AtelyPham (approved):
LGTM 👍

AhmedKorim (approved):

PR #773, opened 2022-12-05
Update Transaction progress for Withdraw providers

### LINKED ISSUE #767: [TASK] Update provider class to handle multiple transactions
## Overview

PR #719 changes the function signature of `deposit` from `Promise<NewNotesTxResult>` to `PromiseLike<NewNotesTxResult>`. This was done to allow custom promise types that can be contained and the transaction state be transferred outside of the provider class enabling the dApp to facilitate multiple transactions. We want to support multiple transactions, as the new bridge dApp flow allows user to begin a new transaction while their existing transaction is being executed. 

Changes have been made in the web3 `vanchor-deposit` to return `Transaction<NewNotesTxResult> ` rather than `Promise<NewNotesTxResult>`. The remaining items for enabling multiple transactions are as follows:

- [ ] Implement the same approach mentioned above by for other providers 
- [ ] Register the executor of the transaction
- [ ] Remove all the notification api calls which fires the toast message and replace it with the `Transaction.next`  or `Transaction.fail` calls

### LINKED ISSUE #768: [TASK] Update the hooks in `useBridgeWithdraw`
## Overview 

Related to the changes mentioned in #767 we need to also update the hooks in `useBridgeWithdraw`. This is a small change and should consist of the following:

- [ ] Remove the transaction state listener for the provider in favour of `useTxQueue`
- [ ] Remove the `setStage` function in favour of deriving the transaction stage from `Transaction.currentStatus`

AhmedKorim:
## Summary of changes
- Updated the abstract provider withdraw return type to be `Transaction<NewNotesTxResult>`
- Removed the transaction state from the abstract provider
- Refactor the withdrawal method of the web3 provider to be shorter 
- Added Transaction support for Polkadot provider
- Remove the `useTransactionStage` hook
- Added tx hash watcher for the `Transaction`
- Updated the `txApiQueue` hook to subscribe to the txHash of the transaction and map it to the `TxPayload`
- Update the Webb interface for the withdraw method to no longer extend the `EventBus`
- Removed the `setStage` calls for `WithdrawContainer`.
- Removed the `setStage` function from `useWithdraw` hook

- Closes #767 
- Close #768 

-----
### Code Checklist 

- [ ] Tested
- [ ] Documented


AtelyPham (approved):
LGTM 👍

PR #785, opened 2022-12-07
Setup the Marketing app in Monorepo

AtelyPham:
## Summary of changes
- Setup a webbsite app for the marketing page in the Monorepo
- Remove warnings about duplicated modules and mismatched version numbers (in `@polkadot/**` libs). Ref [here](https://substrate.stackexchange.com/questions/2399/how-to-prevent-remove-duplicate-modules-and-mismatched-version-numbers-i-e).
- Import the CSS files in apps instead of importing directly in the library `index` file (some apps require a different approach to import CSS files, like `Next.js` and that might cause errors if we import CSS files in the library)
- Add the custom `Identicon` to support SSR. Ref [here](https://github.com/polkadot-js/ui/issues/672#issuecomment-1240125664).

- Closes `NaN`

-----
### Code Checklist 

- [x] Tested
- [x] Documented

### Need further discussion

We're using the `dynamic` function from the `Next.js` lib to support SSR dynamic import in the custom `Identicon` file. This might add more dependencies for our component library, which I think is not really necessary.

Instead of the current approach, I think we can clone a simple version of the `Next.js` `dynamic` library function to support SSR. WDYT? @AhmedKorim @dutterbutter 

### Related links

- The current implementation of the custom `Identicon` file [here](https://github.com/webb-tools/webb-dapp/blob/2ae63d191e0f2527be9f384e695d38dba4361838/libs/webb-ui-components/src/components/Avatar/Identicon.tsx).
- The `Next.js` [dynamic](https://github.com/vercel/next.js/blob/canary/packages/next/shared/lib/dynamic.tsx) function.


dutterbutter:
@AtelyPham I think cloning a simple version of the Next.js dynamic library function to support SSR makes sense here. We should aim to make that component library as lean as possible.

AhmedKorim:
@AtelyPham  please fix the conflicts

dutterbutter (approved):

PR #792, opened 2022-12-08
Update balances watching and handle token selection view not updated

### LINKED ISSUE #717: [BUG] balances don't update for a token on chain switching
## Overview
On the bridge, UI tokens has their balance, if the chain is switched the balance keeps old values

Initially connected to Optimism the input shows the balance for Eth.01
image
switched to Goerli keeps showing .01 for ETH
image
Goerli After a refresh shows the actual balance
image

### LINKED ISSUE #746: [TASK] Enhance`MainComponent` to watch props
## Overview
The `IWebbUIContext` has the `customMainComponent` and `setMainComponent` which are used to set a component dynamically, this component doesn't watch the props passed to it

**Example**

```tsx
  const onConnectWalletClick = useCallback(() => {
    const sourceChains: ChainType[] = Object.values(chains).map((val) => {
      return {
        name: val.name,
        symbol: currenciesConfig[val.nativeCurrencyId].symbol,
      };
    });

    setMainComponent(<ChainSelectionWrapper sourceChains={sourceChains} />);
  }, [chains, setMainComponent]);

```
in this example, the ChainSelectionWrapper wont be notified with the updates on the `chains` object leading to UI bugs

AhmedKorim:
## Summary of changes
- Adding a new hook for listing on balances changes with token address
- Change the chain query return type for balance to be `Observable<string>` instead of `Promise<string>` for both Polkadot and web3 providers.
- Refactor and change the impl the balances related hook/providers to  use the new Observable<string> return type.
- Enhance props for `DepositContainer`
- Realtime balance changes for the token selection list
- Refactor and cleanup DepositContainer 

- Closes #717 
- Closes #746 

### demo

- Balance change from a transfer: https://www.loom.com/share/7dd5cd9aacc94915be359864117020aa
- Balance change from an account switch: https://www.loom.com/share/6ca5028a13174fac9b7d68aa7014a05f
-----
### Code Checklist 

- [x] Tested
- [ ] Documented


libs/polkadot-api-provider/src/webb-provider/chain-query.ts
```diff
@@ -19,59 +20,66 @@ export class PolkadotChainQuery extends ChainQuery<WebbPolkadot> {
...
+            let tokenAccountData: OrmlTokensAccountData;
+
+            try {
+              tokenAccountData = await this.inner.api.query.tokens.accounts(
+                activeAccount.address,
+                assetId
+              );
+            } catch (e) {
+              // It is possible that we have connected to a chain that is not setup with orml tokens.
+              return '';
```
dutterbutter, line 39: Shouldn't we handle the error properly here?

libs/polkadot-api-provider/src/webb-provider/chain-query.ts
```diff
@@ -19,59 +20,66 @@ export class PolkadotChainQuery extends ChainQuery<WebbPolkadot> {
...
+  tokenBalanceByCurrencyId(
     typedChainId: number,
     currency: CurrencyId
-  ): Promise<string> {
+  ): Observable<string> {
     const assetId =
       this.inner.config.currencies[currency].addresses.get(typedChainId);
-    return assetId ? this.getTokenBalanceByAssetId(assetId) : '';
+    if (!assetId) {
+      return throwError(`can't find asset`);
```
dutterbutter, line 76:
```suggestion
      return throwError(`unable to find asset`);
```

dutterbutter, line 76: For error messages we should refrain from using contractions (e.g. can't, shouldn't)

libs/react-hooks/src/currency/useCurrencyAddressBalance.tsx
```diff
@@ -0,0 +1,17 @@
...
+  const currency = useMemo(() => {
+    const currencyConfig = apiConfig.getCurrencyByAddress(address);
+    if (!currencyConfig) {
+      return undefined;
+    }
+    return new Currency(currencyConfig);
+  }, [apiConfig, address]);
+  const balance = useCurrencyBalance(currency);
+  return balance ?? '0';
+}
```
dutterbutter, line 17: This doesn't look like its being used anywhere?

dutterbutter (dismissed):
Where are these files being used? I don't see where they are being called in the code?  

Folder dir in question:
`libs/react-hooks/src/currency/*`

libs/polkadot-api-provider/src/webb-provider/chain-query.ts
AhmedKorim: we can open an issue for better holding of this,but don't think we should do it in this PR

libs/react-hooks/src/currency/useCurrencyAddressBalance.tsx
AhmedKorim: Will remove it, it's no logger being used

apps/bridge-dapp/src/containers/DepositContainer/DepositContainer.tsx
```diff
@@ -537,6 +491,236 @@ export const DepositContainer = forwardRef<
...
+          setDestChain(destChain);
+          setMainComponentName(undefined);
+        },
+      };
+    }, [
+      destChainInputValue,
+      destChains,
+      setDestChain,
+      setMainComponentName,
+      populatedAllTokens,
```
AtelyPham, line 551: `React Hook useMemo has missing dependencies: 'balances', 'bridgeGovernedCurrency', 'chains', 'getPossibleGovernedCurrencies', and 'wrappableCurrency'.`

AtelyPham (approved):
LGTM 👍 Great works 💯  @AhmedKorim

PR #802, opened 2022-12-12
Fetches block time for stats provider from chain

### LINKED ISSUE #801: [TASK] fetch the block time for stats provider from chain
### Overview
 Currently, the `block time` is hardcoded as a provider prop, which isn't accurate it's required for this value to be accurate in order to get the right data from the subquery API.

## Deliverables
- [x] Query the right block target from chain `(await api.consts.timestamp.minimumPeriod) * 2`
- [x] Set this value in the stats provider

devpavan04:
## Summary of changes
- Fetches block time for stats provider from chain

- Closes #801


dutterbutter:
@devpavan04 if this is a WIP let's keep in a DRAFT state please.

drewstone:
@devpavan04 can you ensure this gets wrapped up? What's the delay here?

devpavan04:
### Changes overview

https://user-images.githubusercontent.com/53374218/213242364-342335c7-7a5d-4b76-a67b-d2a25d99088b.mov

PR #1016, opened 2023-02-20
FEAT: Adds Blog to Webbsite with Notion database

devpavan04:
## Summary of changes

- Adds blog landing page
- Adds all blog posts page
- Adds all videos page
- Adds post page

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/stats-dapp`
- [x] `apps/webbsite`

### Screen Recording (First implementation)

https://user-images.githubusercontent.com/53374218/219993868-2dc1dbe9-c7e9-4f9a-b7b9-b5bb1db669b7.mp4

### Screen Recording (Responsiveness)

https://user-images.githubusercontent.com/53374218/219994890-9dab1cd0-f7f7-4a37-a9bd-c21ae30a863e.mp4


devpavan04:
@AtelyPham Because of a few font size and text color discrepancies, I've made changes to mainly `globals.css` file and other files as well.

apps/webbsite/src/libs/notion/notion.ts
```diff
@@ -0,0 +1,160 @@
...
+
+        return {
+          metadata,
+          recordMap,
+        };
+      })
+    );
+
+    return posts;
+  }
```
dutterbutter, line 104: A couple points on this function to consider:

1. Use try-catch block to handle errors: Instead of using a conditional statement to check for errors, you can use try-catch blocks to handle errors in a more efficient way.

2. Add error handling for external service calls: The code is making a call to an external service this.notion.users.retrieve() to fetch the author name. If this call fails, it can cause an unhandled error. Therefore, it's important to add error handling for external service calls.

3. Handle errors for async functions: The Promise.all() function is used to execute multiple async functions concurrently. If any of these async functions reject the promise, it can cause an unhandled error. Therefore, it's important to handle errors for these async functions.

apps/webbsite/src/libs/notion/notion.ts
```diff
@@ -0,0 +1,160 @@
...
+        };
+
+        return {
+          metadata,
+        };
+      })
+    );
+
+    return posts;
+  }
```
dutterbutter, line 137: Other considerations to improved error handling: 

1. Add error handling for the `notion.databases.query` call. You could add a try-catch block to handle any errors that might occur during the query.

2. Handle errors in the `map` function by using try-catch blocks. If an error occurs during the iteration, you can catch the error and return a default value or log the error.

3. Consider using `Promise.allSettled` instead of `Promise.all` to ensure that all promises are settled, even if some of them fail. This will ensure that the function returns an array of objects for all the videos, regardless of whether there were errors during the iteration.

apps/webbsite/src/libs/notion/notion.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    return post;
+  }
+
+  async getPostRecordMap(id: string) {
+    const recordMap = await this.notionAPI.getPage(id);
+
+    return recordMap;
+  }
+}
```
dutterbutter, line 160: A couple points to consider for improved error handling: 

1. Error handling for the getPosts() function: The getPostBySlug function relies on the getPosts() function to retrieve the list of posts. If getPosts() throws an error, it will be unhandled and result in an unhandled rejection. To handle this, we can wrap the getPosts() call in a try-catch block.

2. Type checking for the slug parameter: The slug parameter can be either a string or an array of strings. If an array is passed, the function will not work as intended. We can add a type check to ensure that the parameter is always a string.

3. Improved error message for the posts.find() function: If the posts.find() function returns undefined, the error message will not be helpful. We can improve the error message by checking if post is defined and throwing an error if it is not.

apps/webbsite/src/pages/blog.tsx
```diff
@@ -0,0 +1,53 @@
...
+  videos: Video[];
+};
+
+const Blog: FC<{ blog: BlogProps }> = ({ blog }) => {
+  const { posts, videos } = blog;
+
+  // Featured post is the last one in the array
+  const featuredPost = posts[posts.length - 1];
+  // Recent videos are the last 3 in the array
+  const recentVideos = videos.slice(-3);
```
dutterbutter, line 17: Consider adding checks to ensure that featuredPost and recentVideos have valid values:

```
  const featuredPost = posts.length > 0 ? posts[posts.length - 1] : null;
  const recentVideos = videos.length > 0 ? videos.slice(-3) : [];
```

dutterbutter (changes requested):
A few suggested changes. 

I fixed the build issue with the `.env` variables but it seems you are not hitting a `metadata` undefined issue. When you build locally does this fail as well?

apps/webbsite/src/components/sections/BlogSection.tsx
```diff
@@ -0,0 +1,91 @@
...
+import Link from 'next/link';
+
+type BlogSectionProps = {
+  type: 'post' | 'video';
+  items: PostsOrVideos[];
+  showAllItems?: boolean;
+};
+
+export const BlogSection = (data: BlogSectionProps) => {
+  const { type, items, showAllItems = false } = data;
```
dutterbutter, line 14: Suggestions:

1. Destructure the `data` parameter directly in the function signature instead of inside the function body. This will make the code easier to read and understand.

2. Since `tagsArrays` is only used to compute the `tags` array using the `reduce` function, you can skip the intermediate `tagsArrays` variable and chain the `reduce` function directly on allItems.

3. You can simplify the logic in the `tags` and `filteredItems` memoized values by using the ternary operator and eliminating the else branch.

4. Instead of hard-coding the `BlogCard` component's properties, you can destructure the metadata object directly in the function argument.


```suggestion
export const BlogSection = ({ type, items, showAllItems = false }: BlogSectionProps) => {
```

apps/webbsite/src/components/sections/BlogSection.tsx
```diff
@@ -0,0 +1,91 @@
...
+    return tags;
+  }, [tagsArrays]);
+
+  const filteredItems = useMemo(() => {
+    if (filter === 'All') {
+      return allItems;
+    } else {
+      return allItems.filter((item) => item.metadata.tags.includes(filter));
+    }
+  }, [filter, allItems]);
```
dutterbutter, line 43:
```suggestion
const allItems = showAllItems ? items : items.slice(0, 6);

  const tags = useMemo(() => {
    const tags = allItems.reduce((accumulator, { metadata: { tags } }) => {
      tags.forEach((tag) => {
        if (!accumulator.includes(tag)) {
          accumulator.push(tag);
        }
      });
      return accumulator;
    }, ['All']);

    return tags;
  }, [allItems]);

  const filteredItems = useMemo(() => {
    return filter === 'All'
      ? allItems
      : allItems.filter((item) => item.metadata.tags.includes(filter));
  }, [filter, allItems]);

  const [filter, setFilter] = useState('All');
```

apps/webbsite/src/components/sections/BlogSection.tsx
```diff
@@ -0,0 +1,91 @@
...
+              key={id}
+              title={title}
+              tags={tags}
+              cover={cover}
+              type={type}
+              link={link ? link : `/blog/posts/${slug}`}
+            />
+          );
+        })}
+      </div>
```
dutterbutter, line 80:
```suggestion
      <div className="grid md:grid-cols-2 xl:grid-cols-3 gap-6 mb-[72px]">
        {filteredItems.map(({ id, metadata: { title, slug, tags, cover, link } }) => (
          <BlogCard
            key={id}
            title={title}
            tags={tags}
            cover={cover}
            type={type}
            link={link ? link : `/blog/posts/${slug}`}
          />
        ))}
      </div>
```

dutterbutter:
@devpavan04 it seems the latest commit has caused issues with the netlify build preview... i will look to investigate this but should not be a blocker to have this reviewed. @AtelyPham 

I created this issue #1031 to track improvement to the image loading.

apps/webbsite/src/styles/globals.css
```diff
@@ -57,67 +54,169 @@
...
 }
 
-/** Headings and subheadings */
-
+/* Headings, sub-headings, card texts */
 .heading-1 {
-  @apply text-[32px] leading-[46px] tracking-tight font-bold text-mono-200 font-satoshi
-    md:text-[84px] md:leading-[96px] md:tracking-normal;
+  @apply text-base
+  text-[34px] leading-[45.9px] tracking-[-0.011em] /** small screen */
```
AtelyPham, line 60: You don't need the `text-base` here because you already set the font size by `text-[34px]` and `leading-[45.9px]`. All the `text-base` classes below can be removed.

apps/webbsite/src/styles/globals.css
```diff
@@ -57,67 +54,169 @@
...
 .sub-heading-2 {
-  @apply text-base leading-[22px] font-medium text-mono-180 font-satoshi
-    md:text-[18px] md:leading-[27px];
+  @apply text-base
+  text-[16px] leading-[21.6px] /** small screen */
+  md:leading-[24px] /** medium screen */
+  font-medium font-satoshi;
+}
+
+.button-primary {
```
AtelyPham, line 99: Can you talk more about why you want a custom class name for the primary button and secondary button here, why don't we use the `Button` component from `webb-ui-kit`, the below class is mostly the same with the `Button` component.

apps/webbsite/src/components/sections/ApplicationsAndInfrastructureSection.tsx
```diff
@@ -1,42 +1,37 @@
...
-import SubHeading2 from '../SubHeading2';
+import { Heading2, Heading3, SubHeading1, SubHeading2 } from '../../components';
 
-const ApplicationsAndInfrastructureSection = () => {
+export const ApplicationsAndInfrastructureSection = () => {
   return (
     <section className="py-16 md:py-[156px] space-y-6">
       <div className="pb-9 space-y-9 px-4 max-w-[900px] mx-auto">
-        <Heading2 className="text-center">
+        <Heading2 className="text-center text-mono-200">
```
AtelyPham, line 14: You can set the default color (e.g. `text-mono-200 dark:text-mono-0`) in the `.heading2` class so that you don't need to set the color manually here.

AtelyPham (changes requested):
Some change proposals. 

Do we need this `tools/scripts/local-webb-dapp/protocol-solidity` directory? @devpavan04

apps/webbsite/src/styles/globals.css
devpavan04: @AtelyPham You're right we can just use `Button` component. I'm currently passing `button-primary` and `button-secondary` classname to the `Button` components. We need to override default `Button` styles to match with the design. For example text size, padding is different on different screen sizes in `webbsite` compared to `stats` and `bridge`. Instead of passing overriding classnames to match with the design to all the `Button`'s we're using in `webbsite`, I've created a single class.

Example:

`Button` without overriding class names.

Screenshot 2023-03-02 at 7 23 36 AM

Screenshot 2023-03-02 at 7 15 59 AM

`Button` with `primary-button` class name.

Screenshot 2023-03-02 at 7 25 09 AM

Screenshot 2023-03-02 at 7 16 11 AM

I'm not sure if you can see the subtle difference in the screenshot but there is.

devpavan04:
@dutterbutter @AtelyPham fetches twitter handles from notion.


https://user-images.githubusercontent.com/53374218/222755001-f5371270-38c1-464c-873b-17c381bb81d1.mp4

AtelyPham (approved):
LGTM 👍

Issue #1006, opened 2023-02-15
[TASK] Update localstorage to store an object indexed by create2 address


dutterbutter:
## Overview

The usage of `localStorage` in the bridge causes disruption and failures when new deployments are done. Developers have to clear storage or know exactly what notes are valid for any given deployment. 

We should update how we store encrypted data in localStorage and index the object by create2 address being used in the dApp. It would provide a clear distinction between localStorage objects that can then be removed. We are currently storing by `<NoteAccountPublicKey : {encryptedNotes... }>`. The limitation here is that most developers are making use of the same address to create their Note Account resulting in the appending of notes to this object even though the bridge deployment addresses may have changed. 

### Potential Solution

I potential approach could be to update storage to be <Create2Address: `{NoteAccountPublicKey: {encrypedNotes...}>`. This would provide clear distinction if deployments had changed and reducing the amount of times the dApp fails due to an old notes from a previous deployment. 

Thoughts here @AtelyPham ?

drewstone:
The way I initially envisioned this is to index everything under the specific contract addresses in question:
```
{
	[anchorAddress]: {
		[notePublicKey]: { ...encryptedNotes },
		leaves: [ ...arrayOfLeaves ]
	}
}
```

drewstone:
Scratch my exact thoughts above, instead I think we should be indexing these objects by the `ResourceId` that we use universally around the system. This makes it much easier for us to separate between Substrate / EVM as we are currently planning to handle. The new scheme would then look something like
```json
	[resourceId]: {
		leaves: [ ... ],
		notePublicKey: [ ... ],
	}
```

CC @AtelyPham

Issue #1004, opened 2023-02-14
[BUG] Missing chain selection


drewstone:
## Expected Behavior
In our current 6-sided testnet bridge we should see all testnet bridges as destination options.

### Current Behavior
Instead we see just 5 on the destination side.
<img width="592" alt="image" src="https://user-images.githubusercontent.com/13153687/218676663-2b8b0bfb-1d93-4603-9cd6-e7e7182e3ec0.png">

On the source side we see.
<img width="583" alt="image" src="https://user-images.githubusercontent.com/13153687/218676791-ae7db310-e710-44ae-ac9a-7e1d853e5670.png">

### Possible Solution
However we are getting all valid chains for a given bridge, we need to ensure we are getting all of them consistently.

## Steps to Reproduce

1. Connect to sepolia as source.
2. Click destinations and see only 5 chains.

Issue #233, opened 2021-12-29
[TASK] Remove options with Edgeware and use regular options


drewstone:
https://github.com/webb-tools/webb-dapp/blob/5d05fa04b67f08af6007fda0be8c07eead764640/packages/wallet/src/providers/polkadot/polkadot-provider.ts#L6

PR #1020, opened 2023-02-22
Adding Arbitrum and Indexing localStorage by `ResourceId`

AtelyPham:
## Summary of changes
- Adding Arbitrum anchor
- Indexing localStorage by `ResourceId`

### Proposed area of change

- [x] `apps/bridge-dapp`
- [ ] `apps/stats-dapp`
- [ ] `apps/webbsite`

_Specify any issues that can be closed from these changes (e.g. Closes #233)._
- Closes https://github.com/webb-tools/webb-dapp/issues/1006
- Closes https://github.com/webb-tools/webb-dapp/issues/1004

### Comparing

```
// Current leaves storage                |   // New leaves storage
                                         |
{                                        |  {
  [typedChainId]: {                      |    [resourceId]: {
    [vanchorAddress]: {                  |      lastQueriedBlock: number;
      lastQueriedBlock: number;          |      leaves: string[];
      leaves: string[];                  |    }
    }                                    |  }
  }                                      |
}                                        |
```

```
// Current encrypted notes storage       |   // New encrypted notes storage
                                         |
{                                        |  {
  [accountPubKey]: {                     |    encryptedNotes: {
    encryptedNotes: {                    |      [resourceId]: string[];
      [typedChainId]: string[];          |    }
    }                                    |  }
  }                                      |   
}                                        | 
```

-----
### Code Checklist 


libs/note-manager/src/note-manager.ts
```diff
@@ -30,39 +36,79 @@ export class NoteManager {
...
+    const { chainId, chainType } = parseTypedChainId(+typedChainId);
+
+    const contractAddress = isSource
+      ? note.note.sourceIdentifyingData
+      : note.note.targetIdentifyingData;
+
+    return new ResourceId(contractAddress, chainType, chainId);
+  }
+
   static async initAndDecryptNotes(
```
dutterbutter, line 58: A couple improvement considerations for this function:

1. Use `Object.fromEntries` instead of `reduce` to create the `encryptedNotesMap`.

2. Change the `if` statement that checks if `encryptedNotesMap` is truthy to if `(Object.keys(encryptedNotesMap).length > 0)` to make sure that there are actually encrypted notes.

3. Use `Promise.allSettled` instead of `Promise.all` to make sure that all decryption attempts are completed, even if some fail. This can prevent the function from crashing if there is an error with a single note.

4. Add error handling to the decryption process to catch any errors that may occur during decryption.

libs/polkadot-api-provider/src/webb-provider.ts
```diff
@@ -156,6 +160,11 @@ export class WebbPolkadot
     return this.typedChainId;
   }
 
+  async getResourceId(): Promise<ResourceId | null> {
+    throw new Error('Method not implemented.');
+    return null;
+  }
```
dutterbutter, line 166: Can this be removed if its not implemented?

libs/react-hooks/src/useCurrentResourceId.ts
```diff
@@ -0,0 +1,29 @@
...
+    const fetchResourceId = async () => {
+      const resourceId = await activeApi.getResourceId();
+      setResourceId(resourceId);
+    };
+
+    fetchResourceId();
+  }, [activeApi]);
+
+  return resourceId;
+};
```
dutterbutter, line 29: Consider using the `useCallback` hook to memoize the `fetchResourceId` function to avoid creating a new function reference on every render.

Add error handling to catch and log any errors that occur during the `getResourceId` call.

dutterbutter (changes requested):
Couple initial suggestions to consider.

libs/note-manager/src/note-manager.ts
AtelyPham: The `Object.fromEntries` does not accept a callback for us to add our own logic to filter the notes which are indexed by `ResourceId`. All other comments are resolved as suggested.

libs/polkadot-api-provider/src/webb-provider.ts
AtelyPham: We have to have this method because this method is from the [`WebbApiProvider`](https://github.com/webb-tools/webb-dapp/blob/7950409df0004143e06fd69cfdf45d372c64af86/libs/abstract-api-provider/src/webb-provider.interface.ts#L202) interface that is shared between Web3 API and Polkadot API. In the child class we have to implement all methods from the interface.

libs/react-hooks/src/useCurrentResourceId.ts
AtelyPham: Added the error handling in the `fetchResourceId` function, we cannot use `useCallback` here because the function is inside the `useEffect` hook, and it only runs once the `activeApi` changes.

AtelyPham:
Resolved the comments and gave some feedback. Please review again whenever you have a chance, thanks for your review @dutterbutter.

dutterbutter (approved):

PR #1028, opened 2023-03-01
Adding Currency to Metamask

AtelyPham:
## Summary of changes
- Support adding the currency/token to Metamask when the user clicks the token icon on the `TokenInput` component and `TokenListCard` component

### Proposed area of change

- [x] `apps/bridge-dapp`
- [ ] `apps/stats-dapp`
- [ ] `apps/webbsite`

_Specify any issues that can be closed from these changes (e.g. Closes #233)._
- Closes `NaN`

### Screen Recording

https://user-images.githubusercontent.com/60747384/222197239-9aa1ce38-3cef-4058-b82a-d370d365dfa1.mp4


dutterbutter:
@AtelyPham this is a good start but I think it would be better served if we added a dedicated button to the processing tx cards when a successful deposit is made. That is just an idea we can discuss other options. @monaiuu could you please pick the design up and mock something out. 

I do not think user would intuitively know to hover over the token symbol to add token to metamask. Let's keep this up until @monaiuu can provide some design options.

monaiuu:
Just reviewed the deployment preview. I'm wondering if it's needed for users to add tokens in the deposit and transfer UI since the tokens are not yet received in their wallets. My initial thought is that the 'add token' feature should only be available after the user has received the tokens during withdrawal. Open to hearing your thoughts on this. @AtelyPham @dutterbutter

dutterbutter:
> Just reviewed the deployment preview. I'm wondering if it's needed for users to add tokens in the deposit and transfer UI since the tokens are not yet received in their wallets. My initial thought is that the 'add token' feature should only be available after the user has received the tokens during withdrawal. Open to hearing your thoughts on this. @AtelyPham @dutterbutter

I think that makes sense for the dedicated button but lets keep the ability to add token as its currently implemented in this PR. That is, if you hover the token symbol you can quickly add token.

monaiuu:
> > Just reviewed the deployment preview. I'm wondering if it's needed for users to add tokens in the deposit and transfer UI since the tokens are not yet received in their wallets. My initial thought is that the 'add token' feature should only be available after the user has received the tokens during withdrawal. Open to hearing your thoughts on this. @AtelyPham @dutterbutter
> 
> I think that makes sense for the dedicated button but lets keep the ability to add token as its currently implemented in this PR. That is, if you hover the token symbol you can quickly add token.

An alternative idea is if allowing user to add tokens in the Select Token Card:
Screenshot 2023-03-01 at 2 16 05 PM

We could simply have a link button there when the user hovers over to each token similar to the Chain Selection Card, and the link button display 'add to wallet' in a similar effect if we are unable to render the token amount from their connected wallet. 
Screenshot 2023-03-01 at 2 16 24 PM

Making this proposal because this feature is certainly helpful for first-time users. However, for repeated users to trigger this action directly on the control cards could be a bit bothersome. 

Let me know your thoughts @AtelyPham @dutterbutter

monaiuu:
Just created a prototype for adding tokens to wallet from the Token Select Component. In this flow, the system should be able to recognize whether if user's wallet has the token available in wallet (regardless of balance available). Once the user has added the token to wallet, a balance would be rendered (even if balance is 0). 

Then in all subsequent actions, user would not be prompted to add token to wallet since this action has already been completed.

https://user-images.githubusercontent.com/109539270/222511922-2ff743f9-5033-4cd8-b3c3-cef63bad3727.MOV

Let me know if this is reasonable & implementable. Will proceed with handing off this flow once approved.

AtelyPham:
I'm not sure if we can check whether the user's wallet already has the token or not, I'll explore this and send the feedback soon. The design looks good to me. @monaiuu

dutterbutter:
> I'm not sure if we can check whether the user's wallet already has the token or not, I'll explore this and send the feedback soon. The design looks good to me. @monaiuu

The only api call available to do this is `wallet_watchAsset` as documented [here](https://docs.metamask.io/guide/rpc-api.html#wallet-watchasset). However, based on [this stack exchange](https://ethereum.stackexchange.com/questions/107253/metamask-api-how-to-detect-if-token-already-registered) question I do not think its exactly what we want. 

I am going to approve this PR, and open a subsequent task to track the progression of this feature.

dutterbutter (approved):

dutterbutter:
See https://github.com/webb-tools/webb-dapp/issues/1035

PR #1037, opened 2023-03-04
Remove spend notes from storage

### LINKED ISSUE #1034: [BUG] Note table displays notes that are already spent and deleted
## Issue Description

The delete option in the notes table should delete the given note from storage. I am assuming since we have made changes to how we are storing things in localStorage that we did not reflect that change in the deletion of notes. Currently if you delete a note, it does NOT get deleted from localStorage. 

In addition, when a note gets spent we need to remove it from localStorage. I am assuming this error is caused by the same reason mentioned above. 

### Task Checklist

- [ ] Debug notes table deletion function 
- [ ] Provide a fix to resolve deletion issue
- [ ] Debug notes table displaying notes that are spent
- [ ] Provide a fix to resolve notes table displaying notes that are spent

### Video of note not being deleted, and removing notes that are already spent
https://user-images.githubusercontent.com/29983536/222742222-fa51ca69-1203-408a-9a90-0597015c1999.mp4

devpavan04:
## Summary of changes
- Notes table does not display notes that are already spent or deleted.
- `Clear Data` in the wallet component removes all the notes from local storage.
- Deleting an individual note removes that note from local storage.

### Proposed area of change

- [x] `apps/bridge-dapp`
- [ ] `apps/stats-dapp`
- [ ] `apps/webbsite`

- Closes #1034 

-----
### Code Checklist 


apps/bridge-dapp/src/containers/DeleteNotesModal/DeleteNotesModal.tsx
```diff
@@ -38,13 +38,16 @@ export const DeleteNotesModal: FC<DeleteNotesModalProps> = ({
...
     if (!notes || !noteManager) {
       console.trace('No notes or note manager found');
       return;
     }
 
-    notes.forEach((note) => noteManager.removeNote(note));
+    for (const note of notes) {
+      await noteManager.removeNote(note);
+    }
+
```
AtelyPham, line 50: You can use `Promise.all` or `Promise.allSettled` here for remove note concurrently and consider wrapping the code in the try/catch block for better error handling.

libs/storage/src/storage.ts
```diff
@@ -71,6 +71,12 @@ export class Storage<Store> extends EventBus<{
     return this.data;
   }
 
+  async clear() {
+    this.data = {} as Store;
+    await this.commit(this.name, this.data);
+    this.emit('update', this.data);
```
AtelyPham, line 77: You shouldn’t set the default data to `{}` because some storages have a more complex structure than just `{}` (e.g. `NetworkStore`, `BridgeStorage`). You can add a new method to clear the storage in `StorageHandler` and then call that inside this method. Let the client define that function.

devpavan04: @AtelyPham @dutterbutter I've updated the `clear` method to `reset(data: Store)`

`
 async reset(data: Store) {
    this.data = data;
    await this.commit(this.name, this.data);
    this.emit('update', this.data);
  }
`

We reset the `Note Storage` like this: `await this.noteStorage.reset({})`

The reason behind not including this function in `StorageHandler` is that we do not have access to `data`(this.data) since it is private not accessible outside the class.

libs/note-manager/src/note-manager.ts
```diff
@@ -293,21 +293,25 @@ export class NoteManager {
...
+        promises.push(
+          this.noteStorage.set(chainGroupedNotes[0], encNoteStrings)
+        );
+      }
+
+      await Promise.allSettled(promises);
+    } else {
+      resetNoteStorage();
     }
   }
```
dutterbutter, line 316:
```
async updateStorage() {
  if (this.notesMap.size !== 0) {
    await this.noteStorage.reset({});

    const promises = [...this.notesMap.entries()].map(
      async ([chain, notes]) => {
        const encNoteStrings = notes.map((note) => {
          const noteStr = note.serialize();
          return this.keypair.encrypt(Buffer.from(noteStr));
        });

        await this.noteStorage.set(chain, encNoteStrings);
      }
    );

    await Promise.all(promises);
  } else {
    resetNoteStorage();
  }
}
```

libs/storage/src/storage.ts
```diff
@@ -71,6 +71,12 @@ export class Storage<Store> extends EventBus<{
     return this.data;
   }
 
+  async reset(data: Store) {
+    this.data = data;
+    await this.commit(this.name, this.data);
+    this.emit('update', this.data);
+  }
```
dutterbutter, line 78: Will this commit changes to the store regardless of whether the data has actually changed?

devpavan04: Yes. We're not checking if the data has changed or not in the commit func.

`
async commit(key: string, data: NoteStorage): Promise<void> {
      localStorage.setItem(key, JSON.stringify(data));
    },
`

dutterbutter (approved):

PR #1061, opened 2023-03-21
Tangle website starter

devpavan04:
## Summary of changes
- This PR introduces a new app called `tangle-website` to the monorepo based on the existing `website` app.

## Changes
- Creates a new `tangle-website` app in the `/apps` directory.
- Utilizes repeating components and code from the `website` app.
- Updates monorepo configuration to include `tangle-website`.


apps/tangle-website/src/libs/notion/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './notion';
+export * from './types';
+export * from './server-constants';
```
dutterbutter, line 3: The only Notion files / functions required will be for fetching privacy / terms pages. Everything else can be removed.

apps/tangle-website/src/libs/notion/notion.ts
```diff
@@ -0,0 +1,209 @@
+import { Client } from '@notionhq/client';
```
dutterbutter, line 1: Please remove all mentions of blog related functionality. It will not be needed here.

apps/tangle-website/src/libs/notion/server-constants.ts
```diff
@@ -0,0 +1,20 @@
+export const NOTION_API_KEY = process.env.NOTION_API_KEY;
```
dutterbutter, line 1: Please remove.

apps/tangle-website/src/libs/notion/types.ts
```diff
@@ -0,0 +1,57 @@
+import { ExtendedRecordMap } from 'notion-types';
```
dutterbutter, line 1: Please remove all `consts` that are not privacy or terms.

apps/tangle-website/src/pages/_app.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <WebbUIProvider>
+      <Header />
+
+      <main className="app">
+        <Component {...pageProps} />
+        <Script
+          defer
+          data-domain="webb.tools"
+          src="https://plausible.io/js/script.js"
+        />
```
dutterbutter, line 27: Please remove this script.

dutterbutter (changes requested):
Thanks @devpavan04 ! Can you please remove all the assets from `apps/tangle-website/public` . This would include lottie files, static images, svgs. These will not be present in the Tangle site.

AtelyPham (commented):
Please consider creating a new app using nx commands instead of copying the existing one. If we copy the existing app, we may end up with a lot of redundant files specific to the cloned app that may not be used in the new app. @devpavan04 

### Related links:

- [NX doc.](https://nx.dev/packages/next#creating-applications)
- [Generate app without e2e test directory.](https://github.com/nrwl/nx/issues/8217#issuecomment-998451131)

devpavan04:
Closing in favour of #1071

PR #1071, opened 2023-03-22
Tangle website starter

devpavan04:
## Summary of changes
- This PR introduces a new app called `tangle-website` to the monorepo based on the existing `website` app.

## Changes
- Creates a new `tangle-website` app in the `/apps` directory.
- Utilizes repeating components and code from the `website` app.
- Updates monorepo configuration to include `tangle-website`.


dutterbutter (approved):

apps/tangle-website/src/components/index.ts
AtelyPham, line 1: I think if we shared these components between apps. We should move them to `webb-ui-kit` instead of copying them.

AtelyPham (commented):
We have been doing a lot of copying. If we want to share components and logic instead of copying files, I think we should create a new library.

AtelyPham (approved):
LGTM 👍

PR #1072, opened 2023-03-23
Capture user/transaction errors on Hubble Bridge dApp with Sentry

### LINKED ISSUE #1046: [TASK] Capture user / transaction errors on Hubble Bridge dApp
## Overview

We are using Sentry to detect and track when the dApp crashes, however it does not track when a transaction fails or when a user mistakenly inputs an incorrect value. 

We should capture these errors and failures to better understand where the product needs improvement, and determine what parts of the system need additional reliability. 

We should make use of Sentry.io when we catch an error to report / detect what had failed.

devpavan04:
## Summary of changes

- Captures user / transaction errors on Hubble Bridge dApp with Sentry

### Proposed area of change

- [x] `apps/bridge-dapp`
- [ ] `apps/stats-dapp`
- [ ] `apps/webbsite`

- Closes #1046 

-----
### Code Checklist 


AtelyPham (commented):
In the current implementation, we only capture the deposit error. I believe we should also do this for withdraw and transfer. Can you please confirm? @dutterbutter

dutterbutter:
> In the current implementation, we only capture the deposit error. I believe we should also do this for withdraw and transfer. Can you please confirm? @dutterbutter

Yes we want to capture withdraw and transfer.

AtelyPham (approved):
LGTM 👍 Great work 💯 @devpavan04

PR #1278, opened 2023-05-31
Adds a check for proposal data on proposal detail page to prevent from crashing

devpavan04:
## Summary of changes

- Adds a check for proposal data on proposal detail page to prevent from crashing. This crash is caused as a result of indexing of unknown proposals fix in this [PR](https://github.com/webb-tools/webb-graphql/pull/85) on webb-graphql.
- Currently, when a proposal is added (created later, meaning, created not as a result of `DKGProposalHandlerSection.ProposalAdded` event but created when proposal gets removed i.e as a result of `DKGProposalHandlerSection.ProposalRemoved` event), the event data is missing which causes the crash on stats dapp.
- To prevent this, stats dapp now displays `{expired: true}` in proposal details page when the data is not present and was removed, just like in the tangle event logs shown below.

NOTE: We should further investigate and add a fix to display the missing data. Link to the issue will be added below once created.

Link to issue on `webb-graphql`: https://github.com/webb-tools/webb-graphql/issues/86

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [x] `apps/stats-dapp`
- [ ] `apps/webbsite`
- [ ] `apps/faucet`
- [ ] `apps/tangle-website`

Tangle event logs:

CleanShot 2023-05-30 at 18 04 00

Stats dapp proposal detail:

CleanShot 2023-05-30 at 17 51 35

### DIFF
--- a/apps/stats-dapp/src/containers/ProposalDetail/ProposalDetail.tsx
+++ b/apps/stats-dapp/src/containers/ProposalDetail/ProposalDetail.tsx
@@ -235,7 +235,11 @@ export const ProposalDetail = () => {
             <br />
             <Typography variant="mono1" component="p">
               <ProposalData
-                data={getProposalsData(proposalData.type, proposalData.data)}
+                data={
+                  proposalData.data != '0x00'
+                    ? getProposalsData(proposalData.type, proposalData.data)
+                    : { data: { expired: true } }
+                }
               />
             </Typography>
           </div>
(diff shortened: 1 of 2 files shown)


drewstone:
In the follow-up issue, is there no way we can cross-reference an existing item in the stats dapp with the removal option?

drewstone:
Maybe we can augment the event to show the data if it'll help. Certainly we would like to show _what proposal was removed_ and if the data isn't there that might defeat the transparency of it. What do you think @devpavan04 ?

PR #1371, opened 2023-06-23
Fixes keys table real-time update & progress bar reset

### LINKED ISSUE #1291: [BUG] Stats dApp progress bar looks full when it shouldn't be
## Expected Behavior

Possible bug that the stats dApp progress bar isn't updating properly. This is being written immediately after a rotation but looks like it stalled (according to the stats progress it hasn't reset).

### Current Behavior

When a rotation occurs the progress bar should reset.

### Possible Solution

Identify how the recent changes to pull the session length have affected the progress bar calculation. It definitely worked prior.

## Steps to Reproduce

1. Visit the stats dApp

devpavan04:
## Summary of changes

- Active key container & the progress bar table now updates in real-time when the active key changes.
- Keys table also updates upon rotation without a refresh but there is a few seconds of delay as compared to the active key container & the progress bar which can be improved.

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [x] `apps/stats-dapp`
- [ ] `apps/webbsite`
- [ ] `apps/faucet`
- [ ] `apps/tangle-website`

- Closes #1291 

### Screen Recording

https://github.com/webb-tools/webb-dapp/assets/53374218/8e6a8318-1015-4f9e-a1c5-9f93741b9723


AtelyPham (approved):
LGTM 👍!

PR #1394, opened 2023-07-05
Adds header component to hubble-stats

devpavan04:
## Summary of changes

- Adds initial header component to hubble stats with breadcrumbs and TVL and Volume chips with dummy data.

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [x] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [ ] `apps/webbsite`
- [ ] `apps/faucet`
- [ ] `apps/tangle-website`

### Screenshots

CleanShot 2023-07-05 at 05 28 38


apps/hubble-stats/app/layout.tsx
```diff
@@ -26,13 +27,14 @@ export default function RootLayout({
         <Head>
           <title>Welcome to Hubble Stats!</title>
         </Head>
```
AtelyPham, line 29: We don't need to use the `<Head></Head>` component anymore in the `app` directory. You can export a `metadata` variable to set the [metadata](https://nextjs.org/docs/app/building-your-application/optimizing/metadata) for the page.

apps/hubble-stats/components/Header/Header.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+      {/* TVL and Volume Chips */}
+      <div className="flex items-center gap-4">
+        <Chip color="blue">
+          <BlockIcon size="lg" className="stroke-blue-90 dark:stroke-blue-30" />{' '}
+          TVL: $13,642,124
+        </Chip>
+        <Chip color="blue">
+          <BlockIcon size="lg" className="stroke-blue-90 dark:stroke-blue-30" />{' '}
+          volume: $8,562,122
```
AtelyPham, line 31: You should parameterize the values as component props and assign them a default value, such as `'-'`, for initial rendering.
It would also be beneficial to have a `nice-to-have` prop, such as `isLoading`. This prop can display a spinner to indicate that the data is currently being loaded.

AtelyPham (commented):
Left some comments.

apps/hubble-stats/app/layout.tsx
```diff
@@ -26,13 +27,14 @@ export default function RootLayout({
...
-        <body className="bg-body dark:bg-body_dark bg-cover flex">
+        <body className="bg-body dark:bg-body_dark bg-cover flex h-screen">
           {/* Slide Navigation */}
           <nav className="w-[280px] h-full"></nav>
 
           {/* Main Body */}
           <main className="flex-[1]">
             <div className="max-w-[1000px] mx-auto">
+              <Header />
```
vutuanlinh2k2, line 37: I think you should put this on the page.tsx since each page will have different pages will have different breadcrumb

apps/hubble-stats/components/Header/Header.tsx
```diff
@@ -0,0 +1,36 @@
...
+import {
+  Breadcrumbs,
+  BreadcrumbsItem,
+  Chip,
+} from '@webb-tools/webb-ui-components';
+import Link from 'next/link';
+
+export const Header = () => {
+  return (
+    <div className="flex items-center justify-between pt-6 pb-4">
```
vutuanlinh2k2, line 11: Consider create a breadcrumb props for this Header component to use different breadcrumb for each page

apps/hubble-stats/app/layout.tsx
devpavan04: > I think you should put this on the page.tsx since each page will have different pages will have different breadcrumb

@vutuanlinh2k2 Please take a look at the `Breadcrumbs` implementation in `stats-dapp` [here](https://github.com/webb-tools/webb-dapp/blob/167327419921cf13cc11d0348b590d0fdf86a23c/apps/stats-dapp/src/containers/NavBlocksInfoContainer/NavBlocksInfoContainer.tsx#L52-L106).

Having a breadcrumb component in each page seems to be inefficient since breadcrumbs will always be in the header. We can render breadcrumbs dynamically using the location pathname instead of hardcoding breadcrumb item in each different page. Let me know what you think!

AtelyPham: A pertinent question here is, do we have different headers on different pages? In the design, it appears we have the same header across all pages. The only variance is in the breadcrumbs component, which slightly differs depending on the opening page. Therefore, we should place the header in the root layout to prevent unnecessary re-rendering and re-fetching of data in the header component.

apps/hubble-stats/components/Header/Header.tsx
AtelyPham: Instead of a `breadcrumbs` prop, I believe it would be better if we created a hook named `useBreadcrumbs()`. This hook would return a list of values for the `breadcrumbs` component. We could encapsulate all the breadcrumbs logic and calculations inside this hook. What do you think, @devpavan04 and @vutuanlinh2k2?

AtelyPham (approved):

apps/hubble-stats/components/Header/Header.tsx
devpavan04: Yes this sounds good to me. I can create and improve upon hubble-stats breadcrumbs as we start adding new routes.

PR #1470, opened 2023-07-28
Hubble Stats Breadcrumbs

### LINKED ISSUE #1307: [TASK] Implement breadcrumbs component hubble stats page
## Description
The breadcrumbs will enhance navigation and provide users with clear indications of their current location within the Hubble Stats. This involves Implementing the logic to dynamically generate the breadcrumbs based on the user's navigation path. 

Please use the components from webb UI kit as indicated here: https://webb-tools.github.io/webb-dapp/?path=/docs/design-system-molecules-breadcrumbsitem--docs

Figma Link:
https://www.figma.com/file/KWLQWPgzbMMIj0p5dylcgB/%F0%9F%93%88-Shielded-Pool-Stats?type=design&node-id=814%3A128309&mode=dev

devpavan04:
## Summary of changes

- Adds `Breadcrumbs` to `Header` component in `hubble-stats` dapp.

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [x] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [ ] `apps/webbsite`
- [ ] `apps/faucet`
- [ ] `apps/tangle-website`

- Closes #1307 

### Screen Recording

https://github.com/webb-tools/webb-dapp/assets/53374218/3b393642-5dc8-4d96-977b-655e614f7c57


apps/hubble-stats/app/layout.tsx
```diff
@@ -77,6 +85,35 @@ export default function RootLayout({
...
+        label: activeItem,
+        isLast: true,
+        icon: <CoinIcon />,
+        href: '',
+      });
+    }
+
+    return breadCrumbItems;
+  }, [pathname]);
+
```
AtelyPham, line 116: This component should be the "server component" (or async component to fetch TVL and volume), passed fetched data into the `Header` component to render. Move this logic to the `Header` component, extract it into a React hook, and use it inside the `Header` client component.

AtelyPham: Also, consider moving static variables (`items` and `footer` in this case) outside of the component.

libs/webb-ui-components/src/components/Breadcrumbs/BreadcrumbsItem.tsx
```diff
@@ -36,12 +35,17 @@ export const BreadcrumbsItem = React.forwardRef<
     return (
       <Chip
         color="grey"
```
AtelyPham, line 37: The default props should be defined in the component (`isLast` here), instead of using `BreadcrumbsItem.defaultProps = BreadcrumbsItemsDefaultProps;` as you did. You can read more about this [here](https://react.dev/learn/passing-props-to-a-component#specifying-a-default-value-for-a-prop).

AtelyPham (commented):
Left some comments.

apps/hubble-stats/components/Header/Header.tsx
```diff
@@ -38,16 +37,16 @@ const Header = () => {
...
         />
         <VolumeChip
           color="blue"
           Icon={BlockIcon}
           iconSize="lg"
           iconClassName="stroke-blue-90 dark:stroke-blue-30"
-          value={volume}
-          isLoading={volume ? false : true}
+          value={volumeValue}
+          isLoading={volumeValue ? false : true}
```
vutuanlinh2k2, line 49: `isLoading` should be applied only when we are loading the tvl and volume value, not when the tvl and volume are `undefined`. Maybe you can change the `value` prop here to "-" if those values are `undefined`.

vutuanlinh2k2 (commented):
Left a comment

apps/hubble-stats/app/layout.tsx
```diff
@@ -1,13 +1,15 @@
 'use client';
```
AtelyPham, line 1: You can remove `'use client'` here.

AtelyPham (approved):
Left a comment, otherwise LG 👍

PR #1471, opened 2023-07-31
Add More Rpc Urls to Support Multiple Providers

AtelyPham:
## Summary of changes
- Add more RPC URLs to support multiple providers
- Fix the network storage by moving the `networkSwitched` event to be emitted outside of the if condition
- Fix default Polygon Mumbai RPC URL from `wagmi/references` library is inactive -> Override the an active RPC
- Add multiple provider support for `getViemClient` function.

### Proposed area of change

- [x] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [ ] `apps/webbsite`
- [ ] `apps/faucet`
- [ ] `apps/tangle-website`

### DIFF
--- a/libs/dapp-config/src/chains/evm/index.tsx
+++ b/libs/dapp-config/src/chains/evm/index.tsx
@@ -14,7 +14,7 @@ import {
 } from '@wagmi/chains';
 import { EVMChainId, PresetTypedChainId } from '@webb-tools/dapp-types';
 import { ChainType } from '@webb-tools/sdk-core/typed-chain-id';
-import merge from 'lodash/merge';
+import mergeWith from 'lodash/mergeWith';
 import { DEFAULT_EVM_CURRENCY } from '../../currencies';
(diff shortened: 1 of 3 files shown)


libs/dapp-config/src/chains/evm/index.tsx
```diff
@@ -54,72 +56,147 @@ const localDemeterMulticall3DeploymentBlock = process.env
   ? parseInt(process.env.BRIDGE_DAPP_LOCAL_DEMETER_MULTICALL3_DEPLOYMENT_BLOCK)
   : 0;
 
+// Default rpc url of mumbai is not working so we orverride it
```
drewstone, line 59:
```suggestion
// Default rpc url of mumbai is not working so we override it
```

drewstone (approved):
LGTM

vutuanlinh2k2:
LGTM too!

vutuanlinh2k2 (approved):

PR #1473, opened 2023-07-31
Add Tangle EVM Testnet and update contract addresses

shekohex:
## Summary of changes
- Adds Tangle EVM Testnet (4006)
- Updates the contract addresses deployment
- Updates Faucet with the new token addresses

```
Bridge Addresses: 0x6da94Ec3715C13eAdf2903cc246c768B9AfaF198
Anchor Addresses: 0x19E60bbC8b19CBBA2936BEC4B5CE6542013DF02c
Webb Wrapped tTNT (webbtTNT): 0x6aBC6e8E4E4a1fcB43341CeC49cBd5FA3b494520
```

### Proposed area of change

- [x] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [ ] `apps/webbsite`
- [x] `apps/faucet`
- [ ] `apps/tangle-website`

- Closes None

### Screen Recording

-----
### Code Checklist 

### DIFF
--- a/apps/faucet/src/config/tokens.ts
+++ b/apps/faucet/src/config/tokens.ts
@@ -7,12 +7,15 @@ type TokenConfigType = Record<string, string>;
 const tokens: Record<string, TokenConfigType> = {
   [PresetTypedChainId.AthenaOrbit]: {
     webbWETH: '0x915EbFAd9c4Af2F420782a5F03BD544f9a1FF1d1',
+    webbtTNT: '0x6aBC6e8E4E4a1fcB43341CeC49cBd5FA3b494520',
   },
   [PresetTypedChainId.HermesOrbit]: {
     webbWETH: '0x915EbFAd9c4Af2F420782a5F03BD544f9a1FF1d1',
+    webbtTNT: '0x6aBC6e8E4E4a1fcB43341CeC49cBd5FA3b494520',
   },
   [PresetTypedChainId.DemeterOrbit]: {
     webbWETH: '0x915EbFAd9c4Af2F420782a5F03BD544f9a1FF1d1',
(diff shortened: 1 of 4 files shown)


apps/faucet/src/config/tokens.ts
```diff
@@ -7,12 +7,15 @@ type TokenConfigType = Record<string, string>;
 const tokens: Record<string, TokenConfigType> = {
   [PresetTypedChainId.AthenaOrbit]: {
     webbWETH: '0x915EbFAd9c4Af2F420782a5F03BD544f9a1FF1d1',
```
AtelyPham, line 9: Do we still support `webbWETH` on the faucet? If so, we should update the API to include the token address field. Currently, we only have the `typedChainId` and `walletAddress` fields (details [here](https://www.chatbotui.com/url)). How would the backend know which token to send to the user if we support multiple tokens on the same chain?

AtelyPham (commented):
You should also add the chain data to this file: https://github.com/webb-tools/webb-dapp/blob/develop/libs/dapp-config/src/chains/evm/index.tsx. You can use the [Orbit Hermes](https://github.com/webb-tools/webb-dapp/blob/develop/libs/dapp-config/src/chains/evm/index.tsx#L125-L156) that I've added as a reference. Please let me know if you need any additional information.

apps/faucet/src/config/tokens.ts
shekohex: No, we do not, I did not remove it just in-case it would break something in the UI

AtelyPham: You can remove it. We read and render these values dynamically based on the config. Therefore, if you remove it, that token will not appear on the UI.

AtelyPham (approved):
LGTM 👍!

Issue #1645, opened 2023-09-19
[BUG] Invalid BigNumber String on Tranfer flow


AtelyPham:
The initial discovery and report of this issue was made by @vutuanlinh2k2. In the transfer flow, if you input a decimal point amount (for example, 0.02) and then click on 'transfer' followed by 'confirm transfer', an error will occur.

![image](https://github.com/webb-tools/webb-dapp/assets/60747384/65157cff-1a34-4c5d-9ca4-271a1a401d57)

PR #1648, opened 2023-09-19
Fix Invalid BigNumber String on Tranfer flow

AtelyPham:
## Summary of changes
- Fix invalid `BigNumber` string on the transfer flow
- Remove dest chain on token and pool selection card

### Proposed area of change

- [x] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [ ] `apps/webbsite`
- [ ] `apps/faucet`
- [ ] `apps/tangle-website`

- Closes https://github.com/webb-tools/webb-dapp/issues/1645

### DIFF
--- a/libs/web3-api-provider/src/webb-provider/vanchor-actions.ts
+++ b/libs/web3-api-provider/src/webb-provider/vanchor-actions.ts
@@ -159,8 +159,8 @@ export class Web3VAnchorActions extends VAnchorActions<
         transferUtxo,
         notes,
         feeAmount,
-        refundAmount = ZERO_BIG_INT,
-        refundRecipient = ZERO_ADDRESS,
+        refundAmount,
+        refundRecipient,
       } = payload;
 
       const { inputUtxos, leavesMap } = await this.commitmentsSetup(notes, tx);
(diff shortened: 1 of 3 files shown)


libs/web3-api-provider/src/webb-provider/vanchor-actions.ts
```diff
@@ -171,8 +171,13 @@ export class Web3VAnchorActions extends VAnchorActions<
       // If no relayer is set, then the fee is 0, otherwise it is the fee amount
       const feeVal = relayer === ZERO_ADDRESS ? ZERO_BIG_INT : feeAmount;
       const refund = relayer === ZERO_ADDRESS ? ZERO_BIG_INT : refundAmount;
+
       const recipient =
-        relayer === ZERO_ADDRESS ? ZERO_ADDRESS : refundRecipient;
+        relayer === ZERO_ADDRESS
```
drewstone, line 176: I don't understand why to select the recipient we first check if the relayer is the zero address. As it currently stands the recipient could be set to non-zero, but if relayer is zero, you would set the recipient to zero as well.

drewstone (changes requested):

libs/web3-api-provider/src/webb-provider/vanchor-actions.ts
AtelyPham: When the relayer is a zero address, it implies that users have opted not to use a relayer, thus both the refund amount and the recipient (in this case, the refund recipient address) are also set to zero addresses. This statement can be rephrased as follows:

```ts
const recipient =
  relayer === ZERO_ADDRESS || refundRecipient.length === 0
    ? ZERO_ADDRESS
    : refundRecipient;
```

AtelyPham:
Gonna merge this into develop and QA it again before releasing.

PR #1781, opened 2023-10-19
EVM Staking - Breadcrumbs & Header chips

devpavan04:
## Summary of changes

- [x] Updates layout.
- [x] Adds breadcrumbs component.
- [x] Adds `ERA` & `Session` chips - fetches data from API.

### Screen Recording

https://github.com/webb-tools/webb-dapp/assets/53374218/5026b0b9-9ef9-4277-82f1-a33de66714f9


apps/tangle-dapp/data/HeaderChips/getEra.ts
AtelyPham, line 1: In my opinion, returning `NaN` or `-1` might be more practical than `0` if the API cannot be initialized or an error occurs during data fetching. Personally, I prefer `NaN`.

apps/tangle-dapp/data/HeaderChips/getSession.ts
AtelyPham, line 1: Same here.

libs/webb-ui-components/src/constants/index.ts
```diff
@@ -87,6 +87,10 @@ export const DKG_STATS_KEYS_URL = `${DKG_STATS_URL}/#/keys`;
 export const DKG_STATS_AUTHORITIES_URL = `${DKG_STATS_URL}/#/authorities`;
 export const DKG_STATS_PROPOSALS_URL = `${DKG_STATS_URL}/#/proposals`;
 
+export const POLKADOT_ENDPOINT = 'wss://rpc-archive.tangle.tools';
```
AtelyPham, line 90: Renaming the variable to `TANGLE_RPC_ENDPOINT` would be more sensible. 🤔

AtelyPham (commented):
I've left a few comments. Additionally, I believe we should leverage the RxJS observer pattern to receive data in real-time on the client side. I'm available to assist you with this.

vutuanlinh2k2 (approved):
Other than Tin's comments, LGTM!

PR #1872, opened 2023-12-04
EVM Staking - Nomination Flow

devpavan04:
## Summary of changes

- EVM users can now bond tokens and nominate validators.

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [x] `apps/tangle-dapp`
- [ ] `apps/faucet`

### Screen Recording (UI Implementation)

https://github.com/webb-tools/webb-dapp/assets/53374218/e3bae463-30a2-4e79-ad48-f384ebe13c1a

### DIFF
--- a/apps/tangle-dapp/constants/contract.ts
+++ b/apps/tangle-dapp/constants/contract.ts
@@ -1,3 +1,6 @@
+export const StakinInterfaceAddress =
+  '0x0000000000000000000000000000000000000800';
+
 export const StakinInterfaceABI = [
   {
     inputs: [
--- a/apps/tangle-dapp/constants/evm.ts
+++ b/apps/tangle-dapp/constants/evm.ts
@@ -1,4 +1,4 @@
-import { hexToU8a, u8aToHex } from '@polkadot/util';
+import { u8aToHex } from '@polkadot/util';
 import { decodeAddress } from '@polkadot/util-crypto';
 import { ensureHex } from '@webb-tools/dapp-config';
 import { chainsConfig } from '@webb-tools/dapp-config/chains/chain-config';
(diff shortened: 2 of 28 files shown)


apps/tangle-dapp/constants/contract.ts
```diff
@@ -1,3 +1,6 @@
+export const StakinInterfaceAddress =
```
drewstone, line 1: `StakingInterfacePrecompileAddress`?

apps/tangle-dapp/constants/evm.ts
```diff
@@ -8,76 +8,118 @@ import {
...
-      console.log('nominateTx', nominateTx);
+    const txHash = await evmWalletClient.writeContract(request);
 
-      return nominateTx;
-    }
-  } catch (e) {
-    console.log(e);
+    return txHash;
+  } catch (e: any) {
+    throw new Error(e);
```
drewstone, line 123: This doesn't seem like proper error handling

drewstone: If you're just going to throw an error it's no different than removing the `try/catch` altogether

apps/tangle-dapp/constants/contract.ts
devpavan04: I'll update

apps/tangle-dapp/constants/polkadot.ts
```diff
@@ -142,3 +142,21 @@ export const getValidatorIdentity = async (
     );
   }
 };
+
+export const isNominatorAlreadyBonded = async (
```
drewstone, line 146: Does this really need a try/catch? It seems like you wrap it in a try/catch in the component the calls it.

drewstone (changes requested):
Would be curious about improving the error handling. Seems like try/catches are becoming used repeatedly. It might be overused in some places.

apps/tangle-dapp/constants/evm.ts
devpavan04: @drewstone We catch this error to notify users in this function [here](https://github.com/webb-tools/webb-dapp/blob/71f8eab8a53b4be8d9b0964834e5d09bccf853da/apps/tangle-dapp/containers/DelegateTxContainer/DelegateTxContainer.tsx#L105-L165). But I'll optimize the use of try/catch.

devpavan04:
Closing this since the EVM Staking app will be migrated to a new React app.

PR #1887, opened 2023-12-08
EVM Staking - Nomination Flow, Bond More & Change Reward Destination

devpavan04:
## Summary of changes

- Adds inital `Nomination` flow.
- Adds `Bond More` tokens flow.
- Adds `Change Reward Destination` flow.

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [x] `apps/tangle-dapp`
- [ ] `apps/faucet`

### First-Time Nominator Flow: Bond Tokens, Select Reward Destination, and Nominate Validators.

https://github.com/webb-tools/webb-dapp/assets/53374218/620f27a9-4dcf-4a29-ac6f-325f7f2215d2

### Non-First-Time Nominator Flow: Overwrite Nominations Without Bonding More Tokens, Change Reward Destination, and Bond More Tokens.

https://github.com/webb-tools/webb-dapp/assets/53374218/ca4efc01-87bb-45e5-9dbc-4b1f68ad0377


apps/tangle-dapp/containers/DelegateTxContainer/DelegateTxContainer.tsx
```diff
@@ -0,0 +1,494 @@
...
+            });
+
+            closeModal();
+          }
+        }
+      } else {
+        if (amountToBond > 0) {
+          const bondTxHash = await bondExtraTokens(walletAddress, amountToBond);
+
+          if (bondTxHash) {
```
drewstone, line 224: This is a tremendous amount of nesting and very unclean/unreadable. Can you rework this? We should try to reduce the length of this logic in place of helper functions and less nesting.

apps/tangle-dapp/containers/DelegateTxContainer/DelegateTxContainer.tsx
```diff
@@ -0,0 +1,494 @@
...
+    setPaymentDestination(PaymentDestination.Staked);
+    setSelectedValidators([]);
+    setDelegateTxStep(DelegateTxSteps.BOND_TOKENS);
+  }, [setIsModalOpen]);
+
+  const submitAndSignTx = useCallback(async () => {
+    setIsSubmitAndSignTxLoading(true);
+
+    try {
+      if (isFirstTimeNominator) {
```
drewstone, line 149: Lots of nesting from here. We should keep it to a max of like 2 nested conditionals if possible.

apps/tangle-dapp/containers/UpdatePayeeTxContainer/UpdatePayeeTxContainer.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const submitAndSignTx = useCallback(async () => {
+    setIsUpdatePaymentDestinationTxLoading(true);
+
+    try {
+      const updatePaymentDestinationTxHash = await updatePaymentDestination(
+        walletAddress,
+        paymentDestination
+      );
+
+      if (updatePaymentDestinationTxHash) {
```
drewstone, line 77: If we used the reverse would it reduce our nesting?
```
if (!updatePaymentDestinationTxHash) { // return or handle when doesn't exist }

const updatePaymentDestinationTx = await evmPublicClient.waitForTransactionReceipt({
	hash: updatePaymentDestinationTxHash,
});
...
```

apps/tangle-dapp/data/NominatorStats/usePaymentDestinationSubscription.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  useEffect(() => {
+    let isMounted = true;
+    let sub: Subscription | null = null;
+
+    const subscribeData = async () => {
+      try {
+        const api = await getPolkadotApiRx();
+        if (!api) {
+          throw WebbError.from(WebbErrorCodes.ApiNotReady);
```
drewstone, line 26: We're throwing this but do we catch it when this hook is used? Why do we need this try/catch here then? You're kinda using try/catches everywhere but that's making the code even more unclean.

drewstone (changes requested):
This looks good and the videos are good but we can definitely work to improve code quality. I think you're hugging `try/catch` too aggressively unless you think otherwise (I may be wrong here!).

I see try/catches everywhere however. In the API function calls, in the hooks that use them, and then in the effects/etc. We really only need try/catches at the top level so we can catch and show notification toasts. Otherwise, we should consider letting the awaits fail as usual, especially if they throw their own errors. We would be catching them at the outermost code and try/catch.

devpavan04:
@monaiuu Added search bar.

https://github.com/webb-tools/webb-dapp/assets/53374218/e9e8b4a3-ba44-4541-89a1-e0c393b189af

PR #1902, opened 2023-12-19
Implement `zk-explorer`'s landing page design

### LINKED ISSUE #1824: [FEAT] Implement ZK Ecosystem Explorer Landing Page
## Description
This task is to implement the design of the ZK Ecosystem Explorer landing page, this page should serve as a central hub for users to explore and interact with a curated list of zero-knowledge circuits and projects. Please note that the project should be developed as a separate module, distinct from the main dapp ecosystem, with potential integration into zk.webb.tools.

### Requirements
- [x] **Header Component:** Create a header component to display text and image as depicted in the design.
- [x] **Circuit Cards:** Develop circuit card components to showcase the avatar of the project associated with the circuit, the circuit name, a short blurb of the purpose of the circuit (if possible), and key metrics such as # of stars and constraints. 
- [x] **Project Cards:** Develop project card components to display the project thumbnail image from GitHub (if available), the project name, a brief project description, the # of stars, and contributors, and the # of circuits. 
- [x] **Filter Component:** Implement a dedicated filter and display the number of projects or circuits under each filter category.
- [x] **Search Functionality:** Implement a search bar to filter circuits and projects by keywords.
- [x] **Sorting Dropdown:** Create a sorting feature to organize the list of circuits and projects by popularity, latest addition, and trending status.
- [x] **Counter:** Display real-time counter of total circuits and projects listed on the explorer. 

### Screenshot

Screenshot 2023-12-01 at 12 56 50 PM

yurixander:
## Summary of changes

- Implemented landing page design for `zk-explorer`.
- Added responsiveness and pagination integration.
- Added project filtering logic, which queries the backend once the filters change.
- `Sort by` functionality is also integrated and just waiting for the backend's implementation to handle it.
- Added basic API request functions, preparing for backend API communication.

**Important note**
- The `Input` Webb UI component was extended to allow for the `inputClassName` prop to be passed, as its current `className` prop only applies classes to *a wrapper of the input element*, and not the input itself. This was required in order to make the input rounded, consistent with the design.

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [ ] `apps/tangle-dapp`
- [ ] `apps/faucet`
- [x] `apps/zk-explorer`

- Closes #1824

### Screen Recording

https://github.com/webb-tools/webb-dapp/assets/101931215/f4f1a302-4614-4213-acbd-381750063057

---

### Code Checklist


drewstone:
If you're making an API request each time you click a single filter, it's possible that will greatly overwhelm a server. That seems quite unoptimized. This will also make the user experience potentially suffer if we see everything reload every time you click a new filter.

Another option we should explore if you agree with this logic:
- Gather the first X pages of projects and keep locally, which we will initially filter from.
- As that is exhausted, make API requests.

drewstone:
CC @monaiuu I think the "Submit a project" should be relocated higher up, maybe in the top image but certainly in view instead of hiding down at the bottom.

Also @yurixander when on mobile, the filter options should be collapsed in a hamburger menu and not entirely placed in view automatically.

yurixander:
> If you're making an API request each time you click a single filter, it's possible that will greatly overwhelm a server. That seems quite unoptimized. This will also make the user experience potentially suffer if we see everything reload every time you click a new filter.
> 
> Another option we should explore if you agree with this logic:
> 
>     * Gather the first X pages of projects and keep locally, which we will initially filter from.
> 
>     * As that is exhausted, make API requests.

@drewstone, I thought so as well! It does seem like it could use some optimization. However, there are some considerations that need to be made:

1. **Initial API request to gather all items and then filter locally**
    * **Pros**: Single API request. Good for databases that are not excessively large.
    * **Cons**: Not suitable for large datasets. Data needs to be refreshed, or it can become stale.
    * **Optimization**: Caching strategies like `localStorage` may help offset initial load times on subsequent visits. Streaming the data could also be a great optimization: User gets initial page(s) of items quickly, caches it, and does not need to make frequent susequent API requests.
 
2. **Re-fetch items when constraints change**
    * **Pros**: Always provides most up-to-date information. Constraints apply to *ALL* available projects/circuits on the backend, which is what the user would expect.
    * **Cons**: Higher load on the backend.
    * **Optimizations**: Ratelimiting can be used to prevent abuse. Also, debouncing and batching can be added to constraints to make less API requests.

3. **Gather a limited amount of items, and re-fetch more as needed**
    * **Pros**: Balanced performance. Good for improving initial project load times, and more suited to larger datasets.
    * **Cons**: Filtering, sorting, and constraints are limited to the loaded subset of data, which might not represent all that is available. This could lead to users occassionaly not finding what they're looking for, even though it does actually exist.
    * **Optimizations**: Pagination can be used to load more data as needed.

**Other things to consider**
* I've added debounce time and minimum length for the search query input as a starting point for performance.
* We could batch filter/constraints by having the user trigger the request manually via a button.

yurixander:
> ...
> Also @yurixander when on mobile, the filter options should be collapsed in a hamburger menu and not entirely placed in view automatically.

Sure, I'll get that done.

apps/zk-explorer/components/CardTabs.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+export type CardTabsProps = {
+  selectedTab: ItemType;
+  counts: Record<ItemType, number>;
+  sortByClause: SearchSortByClause;
+  onTabChange: (cardType: ItemType) => void;
+  onSortByClauseChange: (sortByClause: SearchSortByClause) => void;
+};
+
+export const CardTabs: FC<CardTabsProps> = (props) => {
```
vutuanlinh2k2, line 24: consider destructing props here to {selectedTab, onTabChange, ... }

apps/zk-explorer/components/CircuitCard/CircuitCard.tsx
```diff
@@ -0,0 +1,53 @@
+import { Card, Typography } from '@webb-tools/webb-ui-components';
+import { FC } from 'react';
+import Image from 'next/image';
+import assert from 'assert';
+import { StarIcon } from '@radix-ui/react-icons';
+import { CircuitItem } from './types';
+import { ShieldKeyholeLineIcon } from '@webb-tools/icons';
+
+export const CircuitCard: FC<CircuitItem> = (props) => {
```
vutuanlinh2k2, line 9: destruct props here too

apps/zk-explorer/app/page.tsx
```diff
@@ -1,34 +1,235 @@
-import { Card, Typography } from '@webb-tools/webb-ui-components';
-import { ArrowRightUp } from '@webb-tools/icons';
-import { ProjectSubmissionControls } from '../components/ProjectSubmissionControls';
+'use client';
```
vutuanlinh2k2, line 1: find a way of not using 'use client' on page.tsx file

apps/zk-explorer/app/submit/page.tsx
```diff
@@ -0,0 +1,90 @@
+'use client';
```
vutuanlinh2k2, line 1: find a way of not using 'use client' on page.tsx file

apps/zk-explorer/components/HeaderActions.tsx
```diff
@@ -0,0 +1,68 @@
+'use client';
```
vutuanlinh2k2, line 1: u can remove 'use client' here

apps/zk-explorer/components/ItemGrid.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { Button, Typography } from '@webb-tools/webb-ui-components';
+
+export type CardGridProps = PropsOf<'div'> & {
+  selectedItemType: ItemType;
+  projects: ProjectItem[];
+  circuits: CircuitItem[];
+};
+
+export const ItemGrid: FC<CardGridProps> = (props) => {
+  const items =
```
vutuanlinh2k2, line 18: add `useMemo` on both **items** and **wereResultsFound**

apps/zk-explorer/components/LinkCard.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { FC } from 'react';
+import { Card } from '@webb-tools/webb-ui-components';
+
+export type LinkCardProps = PropsOf<'a'> & {
+  href: PageUrl;
+};
+
+export const LinkCard: FC<PropsOf<'a'>> = (props) => {
+  return (
+    <a
```
vutuanlinh2k2, line 12: consider using the Link components provided by Next.js [here](https://nextjs.org/docs/pages/api-reference/components/link) instead of <a> tag

apps/zk-explorer/components/SidebarFilters/SidebarFilters.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+    // Under no circumstances should the `onConstraintsChange` prop
+    // function be called if the constraints themselves have not changed;
+    // therefore, `onConstraintsChange` should not be included in the
+    // effect's dependency array.
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [constraints]);
+
+  const handleConstraintChange = (
```
vutuanlinh2k2, line 145: add useCallback here

apps/zk-explorer/components/SidebarFilters/SidebarFilters.tsx
```diff
@@ -0,0 +1,206 @@
...
+} from './types';
+
+export type SidebarFiltersProps = PropsOf<'div'> & {
+  onConstraintsChange: (constraints: FilterConstraints) => void;
+};
+
+export const SidebarFilters: FC<SidebarFiltersProps> = ({
+  onConstraintsChange,
+}) => {
+  const [constraints, setConstraints] = useState<FilterConstraints>({
```
vutuanlinh2k2, line 19: I'm not sure but this can change to 4 useState hooks to handle each field separately. If you think keeping this is a better option, you can explain it here

apps/zk-explorer/components/SidebarFilters/SidebarFilters.tsx
```diff
@@ -0,0 +1,206 @@
...
+    [FilterCategory.ProofSystem]: [],
+    [FilterCategory.Categories]: [],
+    [FilterCategory.License]: [],
+    [FilterCategory.LanguageOrFramework]: [],
+  });
+
+  // TODO: These are only for testing purposes. Remove these once the actual data is available.
+  const debugProofSystemsOptions: FilterOptionItem[] = [
```
vutuanlinh2k2, line 27: you can add more `useMemo` here and for other fake data below

apps/zk-explorer/components/ProjectCard/ProjectCard.tsx
```diff
@@ -0,0 +1,100 @@
+import { Card, Typography, Chip } from '@webb-tools/webb-ui-components';
+import { FC } from 'react';
+import Image from 'next/image';
+import assert from 'assert';
+import { StarIcon } from '@radix-ui/react-icons';
+import { ProjectItem } from './types';
+
+export const ProjectCard: FC<ProjectItem> = (props) => {
```
vutuanlinh2k2, line 8: destruct props here

vutuanlinh2k2 (changes requested):
Left some comments.

A few notes here:
- Try to utilize `useMemo` and `useCallback` whenever possible for better performance
- You can destruct the props for each React component when declaring the component, this can help you save time not to use **props.something** in the code below

monaiuu (commented):
- The 'Sign in with GitHub' button seems inconsistent with button components in [Webb UI Kit](https://webb-tools.github.io/webb-dapp/?path=/story/design-system-molecules-button--default&globals=theme:dark); missing interactive states.

- Missing `hover` state for [Project Card](https://www.figma.com/file/YdAvVIAmv0VGAigJmAAjAJ/%E2%9A%97%EF%B8%8F--ZK-Explorer?type=design&node-id=638%3A147574&mode=dev) component; added linear gradient to `default` state. 

Screenshot 2023-12-21 at 4 07 45 AM

- When going into mobile breakpoint, keep top nav items to the right bound.
- Menu dropdown component seem to go below texts in header.
- Search bar texts overlap with search icon.
IMG_8434

- To give it more breathing room, update side margins in filter container to `36px`

Screenshot 2023-12-21 at 4 37 02 AM

apps/zk-explorer/components/ItemGrid.tsx
yurixander: For these two, I think it would be best to leave them without memoization, since they are trivial operations (constant time). Although they do get recomputed at every re-render, they are constant time operations so they should have the same performance impact as if I were to `useMemo` with them.

What do you think?

apps/zk-explorer/components/SidebarFilters/SidebarFilters.tsx
yurixander: The function isn't directly passed down as a prop to a child component, should I still proceed to wrap it in `useCallback`?

yurixander:
@monaiuu, let me know what you think; I've implemented the requested design changes. [View here](https://deploy-preview-1902--zkexplorer.netlify.app/)

apps/zk-explorer/components/SidebarFilters/SidebarFilters.tsx
yurixander: The mock data is static and there aren't any complex operations being made. Should I still proceed to use `useMemo` here?

apps/zk-explorer/components/SidebarFilters/SidebarFilters.tsx
yurixander: I was thinking of having them be a single object, since they're always handled together throughout the app; so having 4 would introduce additional complexity, props, and code. What do you think?

apps/zk-explorer/app/page.tsx
vutuanlinh2k2: @yurixander After searching around the Next.js document, one way I found is that you can utilize the [`useSearchParams`](https://nextjs.org/docs/app/api-reference/functions/use-search-params) to manage the search/filter state of the home page. You can try to move the hooks to smaller components with this, making each component client-side without any need to wrap those into a new component. Hope this helps!

apps/zk-explorer/components/ItemGrid.tsx
vutuanlinh2k2: Yes we can keep it this way for now and make new updates when integrating with back-end.

apps/zk-explorer/components/SidebarFilters/SidebarFilters.tsx
vutuanlinh2k2: Yes we can leave this for now and make updates later

apps/zk-explorer/components/SidebarFilters/SidebarFilters.tsx
vutuanlinh2k2: You can handle the state of these fields with `useSearchParams` of Next.js. Check my previous comment [here](https://github.com/webb-tools/webb-dapp/pull/1902#discussion_r1436773680)

apps/zk-explorer/components/Filters/types.ts
vutuanlinh2k2, line 1: not sure these values should be hardcoded or get from the back-end

vutuanlinh2k2 (changes requested):
@yurixander Just one concern here + check this [comment](https://github.com/webb-tools/webb-dapp/pull/1902#discussion_r1436773680)

apps/zk-explorer/app/page.tsx
yurixander: Great find. I'll move to implement those changes. You're right, we should definitely SSR the landing page for best SEO.

apps/zk-explorer/components/Filters/types.ts
yurixander: If we don't have them hardcoded, I'll have to make an initial API call to fetch the filters. It's more flexible, but we'll be relying more on backend API calls.

monaiuu (changes requested):
- [x] Fix top bound signin button & search to be right bound.
Screenshot 2023-12-27 at 1 21 03 PM
- [ ] In mobile view, the signin button & search bar should be above breadcrumbs.
Screenshot 2023-12-27 at 1 21 35 PM

- [ ] Remove outline stroke in this search component in `default` state
Screenshot 2023-12-27 at 1 25 09 PM

apps/zk-explorer/components/ItemGrid.tsx
```diff
@@ -0,0 +1,62 @@
...
+  circuits: CircuitItem[];
+};
+
+export const ItemGrid: FC<CardGridProps> = ({
+  selectedItemType,
+  projects,
+  circuits,
+  ...rest
+}) => {
+  const items = selectedItemType === ItemType.Project ? projects : circuits;
```
vutuanlinh2k2, line 24: add **useMemo**

apps/zk-explorer/components/SearchInput.tsx
```diff
@@ -0,0 +1,116 @@
...
+    } else {
+      setSearchParam(SearchParamKey.SearchQuery, newSearchQuery);
+
+      if (onValueChange !== undefined) {
+        onValueChange(newSearchQuery);
+      }
+    }
+  };
+
+  const searchQueryPlaceholder =
```
vutuanlinh2k2, line 93: add **useMemo**

apps/zk-explorer/components/Filters/Filters.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  // Fetch categories and options from API.
+  useEffect(() => {
+    fetchFilterOptions()
+      .then((responseData) => setCategories(responseData.categories))
+      // TODO: Temporarily use mock data until we have a backend.
+      .catch(() => setCategories(MOCK_CATEGORIES));
+  }, []);
+
+  const handleOptionCheckboxChange = (
```
vutuanlinh2k2, line 45: add **useCallback**

apps/zk-explorer/components/HeaderActions.tsx
```diff
@@ -0,0 +1,95 @@
...
+  doHideSearchBar,
+  ...rest
+}) => {
+  // TODO: Should throw error when client id is empty.
+  const githubOAuthClientId = process.env.ZK_EXPLORER_GITHUB_CLIENT_ID ?? '';
+
+  const breakpoint = useTailwindBreakpoint();
+  const { setSidebarOpen, updateSidebarContent } = useSidebarContext();
+
+  const prepareAndShowSearchSidebar = () => {
```
vutuanlinh2k2, line 37: add **useCallback**

vutuanlinh2k2 (approved):

PR #1942, opened 2024-01-11
feat(zk-explorer): Implement design and functionality of Proof Generation page

### LINKED ISSUE #1831: [FEAT] Implement Proof Generation Initiation Flow
## Description
This task is to implement the proof generation initiation flow. The interface should guide users through the steps necessary to upload circuit parameters and select service options. The service is initiated from the [project overview page](https://github.com/webb-tools/webb-dapp/issues/1826) for applicable circuits (`circom`). User is required to have an [account](https://github.com/webb-tools/webb-dapp/issues/1823) prior to going through the flow. 

### Initiation Steps

1. **Process R1CS File:** Implement upload component to allow users to upload their R1CS file. The system should verify the format and integrity of the file and provide immediate visual confirmation of a successful upload or detailed error messages if the upload fails. The system also needs to check whether if the uploaded file matches the circuit selected from the circuit overview page.

2. **Upload Verification Key:** Enable users to upload the Verification Key. The interface must accept .json or .vk files and should validate the file against expected formats. The system should automatically cross-reference the key with the previously uploaded R1CS file to ensure they correspond to the same circuit setup. 

3. **Upload Proving Key:** Similar to above, this step requires the user to provide the Proving Key.
4. **Select MPC Participants:** Display a list of available MPC participants. Users should be able to browse, filter, and select from this list. Each participant's entry should include their identity, location, slashing incidents, and uptime for past performances. By default, all MPC participants are selected, user can deselect specific participants based on preferences. 
5. **Select Servier Tiers:** Allow users to select a tier and clearly display the benefits associated with their choice. 

### Design Specification
- The stepper component should update to reflect the current step of the process. Implement conditional logic to ensure users can only proceed if the current step is completed.
- Implement feedback component that allows users to report issues or suggestions about the initiation process.

### Screenshot
Screenshot 2023-11-14 at 3 41 20 PM

yurixander:
## Summary of changes

- Implemented design of the Proof Generation Initiation page
- Made design responsive and tested light theme
- Added basic frontend functionality, such as ensuring that the user can only move on to the next step after the current step is complete
- Feedback card now links to creating a GitHub issue (which has other useful links, such as feedback, etc.)
- API errors are now handled more gracefully to improve user experience
- `Run Service` modal and design were implemented on project details page

[Click here to view the page on the deployment preview](https://deploy-preview-1942--zkexplorer.netlify.app/project/webb/masp/pairing.circom)

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [ ] `apps/tangle-dapp`
- [ ] `apps/faucet`
- [x] `apps/zk-explorer`

- Closes #1831

### Screen Recording

https://github.com/webb-tools/webb-dapp/assets/101931215/fe8a1a80-9e3d-4d5c-b58b-22c032bd75a0

---

### Code Checklist


apps/zk-explorer/components/LinkCard.tsx
```diff
@@ -3,28 +3,35 @@ import { PropsOf } from '@webb-tools/webb-ui-components/types';
...
-        'block hover:translate-y-[-6px] transition duration-100',
-        className
-      )}
-    >
-      <Card className="p-6 shadow-xl items-start space-y-0">{children}</Card>
+  const content = (
+    <Card className="p-6 shadow-xl items-start space-y-0">{children}</Card>
+  );
+
+  const finalClassName = twMerge(
```
vutuanlinh2k2, line 23: Wrap useMemo here

apps/zk-explorer/components/RadioCard.tsx
```diff
@@ -0,0 +1,79 @@
...
+export const RadioCard: FC<RadioCardProps> = ({
+  id,
+  setSelectedRadioItem,
+  selectedRadioItem,
+  title,
+  tooltip,
+  description,
+  Icon,
+}) => {
+  const isChecked = selectedRadioItem === id;
```
vutuanlinh2k2, line 26: Add useMemo

apps/zk-explorer/components/RadioCard.tsx
```diff
@@ -0,0 +1,79 @@
...
+    {
+      if (selectedRadioItem === id) {
+        return;
+      }
+
+      setSelectedRadioItem(id);
+    }
+  }, [id, selectedRadioItem, setSelectedRadioItem]);
+
+  const isCheckedClass = isChecked
```
vutuanlinh2k2, line 38: Add useMemo

apps/zk-explorer/components/RadioCard.tsx
yurixander: The comparison is a constant time operation, so adding `useMemo` here might introduce overhead. Should I still proceed with using it?

apps/zk-explorer/components/StepTrackerSidebar.tsx
```diff
@@ -0,0 +1,145 @@
+'use client';
```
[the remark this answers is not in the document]
yurixander: Good catch! I had it with `useState` before, and it was left out.

apps/zk-explorer/components/RadioCard.tsx
yurixander: The comparison is a constant time operation, so adding useMemo here might introduce overhead.

I did something a bit different however: Added `useMemo` for the `twMerge` calls, which is likely O(n). Going to start doing that everywhere I use `twMerge` that involves an input prop of unknown length.

apps/zk-explorer/components/RadioCard.tsx
vutuanlinh2k2: I don't really understand the **constant time operation** part, can you elaborate more on that?

yurixander: My thought process was that comparisons (say equality comparisons) in JavaScript and most languages are fast and efficient operations (aka O(1) worst-case).

So, using `useMemo` (whose primary purpose is to optimize by memoizing complex, computationally heavy operations) would not provide a benefit and perhaps even add overhead (function call vs. simple comparison operation).

Every time the component is re-rendered, the variable is re-computed. But that wouldn't mean that anything depending on that variable would *re-render* as well since React only re-renders when the variable's value *actually* changes.

For operations that process arrays linearly (like `for every element in X, do Y`), more complex data structures, or computationally heavy calculations, `useMemo` would be more appropriate.

vutuanlinh2k2 (approved):

apps/zk-explorer/constants/index.ts
```diff
@@ -4,5 +4,11 @@ export const GITHUB_LANGUAGE_COLORS_API_URL =
 export const API_PREFIX = '/api';
 export const ITEMS_PER_PAGE = 12;
 
+export const FEEDBACK_URL =
+  'https://github.com/webb-tools/webb-dapp/issues/new/choose';
```
AtelyPham, line 8: We already have this constant in the `libs/webb-ui-components/src/constants/index.ts:77`.

apps/zk-explorer/utils/utils.ts
```diff
@@ -185,3 +149,19 @@ export function computeUserDiff(initial: User, updated: User): Partial<User> {
...
+  } catch (possibleError) {
+    if (possibleError instanceof Error) {
+      return possibleError;
+    }
+
+    return new Error(
+      'Unknown error because the thrown object is not an instance of an error'
+    );
+  }
+}
```
AtelyPham, line 167: I believe we already have a similar function called `safeParseJson` in other dApps. It might be beneficial to consolidate these functions into a single one and place it in the `webb-ui-components`. This way, we can reuse it across various dApps, maintaining a single source of truth for the project.

AtelyPham (commented):
Left some comments.

apps/zk-explorer/utils/utils.ts
yurixander: Do note that this will involve changes to `bridge-dapp`. Will need to test it somehow, to ensure nothing broke. I've made as few changes as possible to `bridge-dapp` to have it accustomed to the new consolidated `safeParseJson` function.

AtelyPham: Yep, sounds good to me 🆗

AtelyPham (approved):

PR #1966, opened 2024-01-18
feat(tangle-dapp): Payouts Table & Payout Flow

### LINKED ISSUE #1671: [CHECKLIST] EVM Staking UI
# Overview
The EVM users do not have a UI to interact with the staking system. This would be the first step to building it. It should be fairly straightforward. We need to:
- Display the list of validators similar to Polkadot Apps.
- Display a nomination page (also called "Accounts" on Polkadot Apps)
- Allow users to click create nominator accounts (and stash accounts generally).
- Allow users to select validators to nominate and bond funds.
- Execute all transactions through the EVM. 

The implementor of the tasks in this checklist MUST go through the flows on Polkadot Apps. If you haven't done so before, **_please do that_**. It's pointless to have you implement without spending the 5 minutes necessary to actually use the existing functionality so you get a feel for it.

## Example UIs
- https://polkadot.js.org/apps/?rpc=wss://rpc.tangle.tools#/staking
- https://www.dtmb.xyz/moonbeam/collators

## Checklist
- [x] #1677 
- [x] #1776
- [x] #1678 
- [x] #1766
- [x] #1764
- [x] #1763
- [x] #1680
- [x] #1679  
- [x] https://github.com/webb-tools/webb-dapp/issues/1895
- [x] https://github.com/webb-tools/webb-dapp/issues/1900
- [x] #1681
- [x] https://github.com/webb-tools/webb-dapp/issues/1896
- [x] https://github.com/webb-tools/webb-dapp/issues/1910
- [ ]  https://github.com/webb-tools/webb-dapp/issues/1907
- [x] #1926
- [x] #1927

### Thoughts on components to add and reusability.
The components here should mostly exist in our library. Some things that might not exist are the validator selection in the nominator flow, which is like a 2 panel multi-selection.

Otherwise, there is a lot of redundancy. Every action on Apps UI usually ends with an "authorize transaction"/"submit transaction" panel, we could add this but its not totally necessary (we can go directly to submission).

### LINKED ISSUE #1907: [TASK] Implement staking reward payout flow
## Description
Implement staking rewards payout flow, enabling users to claim their staking rewards from validators for each staking era.

Users select the relevant era(s)/validator and initiate the claim process.This transaction includes a transaction fee, paid by the user initiating the claim. Once the transaction is processed, the disbursed rewards are split between the validator and their nominators, minus the validator's commission.

The rewards are directly deposited into each participant's account, and should be displayed in user's total rewards. 

Note: we will only display unclaimed rewards, no need for displaying claimed rewards in the payout table. 

### Screenshot
Image

devpavan04:
## Summary of changes

- Users can view the payouts they have received in the latest history depth. The table displays the era number, validator address, validator's total reward for the era, and the user's reward.

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [x] `apps/tangle-dapp`
- [ ] `apps/faucet`
- [ ] `apps/zk-explorer`

- Closes #1907 
- Closes #1671 

### Screen Recording (Payouts Table)

https://github.com/webb-tools/webb-dapp/assets/53374218/82d9ae86-54b0-4632-b699-8be137262454

### Screen Recording - Payouts Flow - Payment Destination is Staked (auto-compounding)

https://github.com/webb-tools/webb-dapp/assets/53374218/776a7f2c-325e-46d6-9d6c-90790b6624c2

https://github.com/webb-tools/webb-dapp/assets/53374218/1b503686-ed4d-4919-b858-80a06dc150f0

### Screen Recording - Payouts Flow - Payment Destination is Stash (no auto-compounding)

https://github.com/webb-tools/webb-dapp/assets/53374218/c7a3242b-5db4-419a-850a-cea8505e411d


drewstone:
Merge conflicts here

devpavan04:
### Payout and Payout All Flow


https://github.com/webb-tools/webb-dapp/assets/53374218/118b130a-eaed-4cc0-93b7-320eda81e73a

apps/tangle-dapp/data/DelegationsPayouts/usePayouts.ts
```diff
@@ -0,0 +1,288 @@
...
+                        typeof rewards.individual === 'object' &&
+                        rewards.individual !== null
+                      ) {
+                        Object.entries(rewards.individual).forEach(
+                          ([key, value]) => {
+                            if (key === validator.toString()) {
+                              validatorRewardPoints = Number(value);
+                            }
+                          }
+                        );
```
yurixander, line 93: Please correct me if I am understanding wrong, I'm assuming that we're looking for a single specific entry here, and if/once we find it, we set the `validatorRewardPoints` to its value.

In that context, could convert this to a `for..of` loop, and then `break` when you find the `key === validator.toString()`, that way the iteration doesn't continue when not needed:

```ts
for (const [key, value] of Object.entries(rewards.individual)) {
  if (key === validator.toString()) {
    validatorRewardPoints = Number(value);
    
    // No need to continue since we have found the validator.
    break;
  }
}
```

Regardless, if it's an object and we only apply the assignment for a specific key, can't we directly access the value `rewards.individual[validator.toString()]` or am I missing something?

apps/tangle-dapp/data/DelegationsPayouts/usePayouts.ts
```diff
@@ -0,0 +1,288 @@
...
+                    if (claimedEras.includes(era)) return;
+
+                    const erasTotalReward =
+                      await apiPromise.query.staking.erasValidatorReward(era);
+
+                    if (erasTotalReward.isSome) {
+                      const validatorTotalReward =
+                        (validatorRewardPoints *
+                          Number(erasTotalReward.unwrap().toString())) /
+                        totalRewardPoints;
```
yurixander, line 134: Is `totalRewardPoints` always guaranteed to never be `0` here? Otherwise we may have division by zero

apps/tangle-dapp/data/DelegationsPayouts/usePayouts.ts
```diff
@@ -0,0 +1,288 @@
...
+                            validator
+                          );
+
+                        const validatorTotalStake = eraStaker.total.unwrap();
+
+                        const validatorTotalStakeFormatted =
+                          await formatTokenBalance(validatorTotalStake);
+
+                        if (
+                          Number(validatorTotalStake.toString()) > 0 &&
```
yurixander, line 157: This conversion may fail if `Number` can't hold larger Big Integer (`BN`)? If it is a `BN`, it should provide you with functions to accomplish comparisons, ex. `validatorTotalStake.gt(new BN(0))`

apps/tangle-dapp/data/DelegationsPayouts/usePayouts.ts
```diff
@@ -0,0 +1,288 @@
...
+                          );
+
+                          if (
+                            nominatorStakeInfo &&
+                            !nominatorStakeInfo.isEmpty
+                          ) {
+                            const nominatorTotalStake =
+                              nominatorStakeInfo.value.unwrap();
+
+                            if (Number(nominatorTotalStake.toString()) > 0) {
```
yurixander, line 171: If `nominatorTotalStake` is of type `BN` or a derivative, prefer using `.gt(new BN(0))`, since this conversion may fail

apps/tangle-dapp/data/DelegationsPayouts/usePayouts.ts
```diff
@@ -0,0 +1,288 @@
...
+                              const nominatorStakePercentage =
+                                (Number(nominatorTotalStake.toString()) /
+                                  Number(validatorTotalStake.toString())) *
```
yurixander, line 174: If these values are of type `BN` or a derivative, prefer using `.gt(new BN(0))`, since this conversion may fail

yurixander: Must also ensure that `validatorTotalStake` can never be `0` here. But if that's assured, then disregard this comment!

apps/tangle-dapp/data/DelegationsPayouts/usePayouts.ts
```diff
@@ -0,0 +1,288 @@
...
+
+              if (isMounted) {
+                const validatorPayouts = await Promise.all(
+                  validatorPayoutsPromises
+                );
+
+                setPayouts(
+                  validatorPayouts
+                    .filter((payout) => payout !== undefined)
+                    .sort((a, b) => Number(a.era) - Number(b.era))
```
yurixander, line 258: If `era` is a `BN` or its derivative, this conversion may fail for larger values of `BN`. Any other way of doing this sorting comparison?

apps/tangle-dapp/utils/convertToSubstrateAddress.ts
```diff
@@ -1,25 +1,15 @@
...
-export const convertToSubstrateAddress = (
-  address: string,
-  substratePrefix: number = SS58_PREFIX
-): string => {
+export const convertToSubstrateAddress = (address: string): string => {
   if (isEthereumAddress(address)) {
-    return evmToAddress(address, substratePrefix);
+    return evmToAddress(address);
+  } else {
+    return 'invalid';
```
yurixander, line 13: Should we throw an error here instead OR `return null` (to force the caller to handle the possibility that this may fail), to avoid having the caller continue execution with an invalid address?

libs/webb-ui-components/src/constants/index.ts
```diff
@@ -99,7 +99,7 @@ export const DKG_STATS_KEYS_URL = `${DKG_STATS_URL}/#/keys`;
 export const DKG_STATS_AUTHORITIES_URL = `${DKG_STATS_URL}/#/authorities`;
 export const DKG_STATS_PROPOSALS_URL = `${DKG_STATS_URL}/#/proposals`;
 
-export const TANGLE_RPC_ENDPOINT = 'wss://testnet-rpc.tangle.tools';
+export const TANGLE_RPC_ENDPOINT = 'http://localhost:9944';
```
yurixander, line 102: Won't this make the Netlify deployment previews or anything that depends on this constant attempt to connect to localhost instead of the testnet?

yurixander (changes requested):
Great job on this @devpavan04 👍🏼 I can tell it was a difficult task! I've left some suggestions that may help in regards of error handling or refactoring.

apps/tangle-dapp/data/DelegationsPayouts/usePayouts.ts
devpavan04: It'll never be zero.

drewstone:
CI failing @devpavan04

drewstone:
Merge conflicts after recent merge too @devpavan04

PR #2028, opened 2024-02-13
feat(tangle-dapp): Add transaction confirmation cards

devpavan04:
## Summary of changes

- Adds transaction confirmation cards to all staking functionalities.
- Confirmation cards displays status of the transaction and a button which takes user to the block explorer.

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [x] `apps/tangle-dapp`
- [ ] `apps/faucet`
- [ ] `apps/zk-explorer`


apps/tangle-dapp/components/TxnConfirmationCard/TxnConfirmationCard.tsx
vutuanlinh2k2, line 1: I think TxnConfirmationModal is a better name here

vutuanlinh2k2, line 1: I think this approach is more efficient

- Put the `TxnConfirmationModal` in the Layout instead of each file that executes transactions
- Create a hook to handle the state (open, hash, etc.) 
- import the hook to use in components that execute tx

In this way, we can significantly reduce code size. 

Let me know what you think @devpavan04

vutuanlinh2k2 (commented):
Left some comments

apps/tangle-dapp/containers/Layout/Layout.tsx
```diff
@@ -37,6 +38,8 @@ const Layout: FC<PropsWithChildren> = ({ children }) => {
         <Footer isMinimal className="py-8" />
       </main>
 
+      <TxnConfirmationContainer />
```
vutuanlinh2k2, line 41: I think `TxnConfirmationModalContainer` is a better name to let others know this is a modal

apps/tangle-dapp/containers/TxnConfirmationContainer/TxnConfirmationContainer.tsx
```diff
@@ -0,0 +1,21 @@
+'use client';
+
+import { TxnConfirmationModal } from '../../components/TxnConfirmationCard';
+import { useTxnConfirmation } from '../../context/TxnConfirmationContext';
+
+export const TxnConfirmationContainer = () => {
+  const { txnConfirmationState, setTxnConfirmationState } =
+    useTxnConfirmation();
```
vutuanlinh2k2, line 8: `useTxnConfirmationModal` might be a better name here

apps/tangle-dapp/components/TxnConfirmationCard/TxnConfirmationCard.tsx
vutuanlinh2k2, line 1: You forget to change the file name here

vutuanlinh2k2, line 1: I think we should set all the variable using **txn** to just **tx**. **Example**: `TxnConfirmation` to `TxConfirmation` since we are using a lot **tx** as variable name already

vutuanlinh2k2 (commented):
Left some comments

vutuanlinh2k2 (approved):

PR #2029, opened 2024-02-13
[FEAT] Lazily load validator table on EVM staking, Tangle dapp

### LINKED ISSUE #2013: [BUG] Lazily load validator tables on EVM staking, Tangle dapp
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The active validator list must be lazily loaded on the Tangle app EVM staking page, since it is causing performance issues. 

This is because, currently, for every validator, multiple data points are fetched: commission, identity, self-staked amount, etc. Since there are 600 validators, 600*5 = **~3,000** requests and subscriptions are made on each page load.

The ideal way to handle this would be to load validators using pagination lazily.

This performance improvement may also apply to other things, such as [`getWaitingValidators`](https://github.com/webb-tools/webb-dapp/blob/266f784f4fe5a938343ef5731b4fd8398213dd81/apps/tangle-dapp/data/ValidatorTables/getWaitingValidators.ts).

[See relevant code here](https://github.com/webb-tools/webb-dapp/blob/266f784f4fe5a938343ef5731b4fd8398213dd81/apps/tangle-dapp/data/ValidatorTables/getActiveValidators.ts)

### Expected Behavior

* The active & waiting validator tables should lazily load data on page load.

### Steps To Reproduce

_No response_

### Screenshot

image

### Anything else?

_No response_

yurixander:
## Summary of changes

- Lazy load validator table through pagination. This avoids fetching all 750+ validators on page load.
- Apart from lazy loading, local storage caching is also used to ensure that the user perceives a significantly faster loading time, at least on the validator table.
- Refactored and merged the common logic in `getActiveValidators` and `getWaitingValidators` into one function (`fetchValidatorDetails`) to avoid repeating code.
- Created re-usable `usePagedValidators` hook, to fetch a certain amount of validators given a page and page size, which allows it to be easily integrated with pagination for tables & other components in a lazy way.
- Added development-only request & subscription counter toolbar, useful for optimizing Polkadot API request count, and seeing if there's large amounts of requests being made while debugging (see attached preview below).

Development only request counter toolbar

### Other points that need optimization
- [useActiveAndDelegationCountSubscription](https://github.com/webb-tools/webb-dapp/pull/2029/files#diff-8f0ce52c9ea8bdfc470968e68bde13b40c2a4be31613963a25e0259755696c8e): This is fetching information **per** validator. Only solution to reduce request count seems to be batching, if even possible in this case.
- [getActiveValidators](https://github.com/webb-tools/webb-dapp/pull/2029/files#diff-3248ce9159ddadb3f51280bae79cbc16cb2793d17c484e68af255682123496c8): This is fetching ~5 data points **per** validator. Only solution to reduce request count seems to be batching, if even possible in this case.
- [getWaitingValidators](https://github.com/webb-tools/webb-dapp/pull/2029/files#diff-35c891e1e99f4d1774358ef02ba790d3d3314255713ba35219839bbb9fe00eb6): Same as `getActiveValidators`.

### Additional notes
- The current implementation of some of the components (ie. nomination list) seems to force us to perform at least as many requests as there are validators (750 as of now), and then 5 data points per each validator.
- Some solutions to this that I've both considered & attempted:
  - __Web workers__: To prevent heavy async data fetching from stalling UI. It seems it can't easily be implemented in Next.js. Regardless, this does not reduce the total amount of requests made.
  - __Caching__: Not ideal to cache real-time data, even if I did, the validator list would need to be refreshed periodically either way, which would still stall the UI.
  - __Server-side data fetching__: Would not work well given that this is real-time data, and we want it updated. Server-side props would just defer the data fetching till later, otherwise the visitor would be served with data that goes stale and doesn't update.
  - __Batching requests using Polkadot's multi-query__: Tried to go this route, but from what I've seen so far, it seems like it may be overly complex to implement for this specific case ([not necessarily querying related data, as Polkadot API might expect](https://polkadot.js.org/docs/api/start/api.query.multi/)).
  - If batching cannot be implemented for fetching validator details, we may need to consider the following options:
    1. Leave request amount as-is, find a way to prevent UI from stalling. This should be possible given that Polkadot Portal seems to be managing to accomplish this (See `Network -> Staking`).
    2. Consider using GraphQL backend to have an endpoint which would return exactly the data needed here, given a single request.

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [x] `apps/tangle-dapp`
- [ ] `apps/faucet`
- [ ] `apps/zk-explorer`

- Closes #2013

### Screen Recording

https://github.com/webb-tools/webb-dapp/assets/101931215/457d07f0-dd63-43e8-96eb-3efb1103e067

---

### Code Checklist


apps/tangle-dapp/containers/ApiDevStatsContainer/ApiDevStatsContainer.tsx
```diff
@@ -0,0 +1,151 @@
+'use client';
+
+import { Expand } from '@webb-tools/icons';
+import { SkeletonLoader, Typography } from '@webb-tools/webb-ui-components';
+import { FC, useEffect, useState } from 'react';
+
+import usePromise from '../../hooks/usePromise';
+import { getPolkadotApiPromise, getPolkadotApiRx } from '../../utils/polkadot';
+
+const MEGABYTE_FACTOR = 0.000001;
```
vutuanlinh2k2, line 10: can you add comments to explain the `MEGABYTE_FACTOR` variable and the `formatBytes` function here

vutuanlinh2k2 (changes requested):
left a comment

vutuanlinh2k2 (approved):

Issue #1991, opened 2024-01-23
[TASK] Independent Profile Update Flow


monaiuu:
## Description
Users who have initially set up their profile with independent restaking need the ability to update their profile. This task involves creating a user interface and logic to adjust the restake amount for each role and to add or remove roles, taking into account the activity status of each service role.

### Key Points:
![Image](https://github.com/webb-tools/webb-dapp/assets/109539270/0849f9c4-72ec-4319-84eb-0845d76a0508)
- Increase Stake for Any Role: 
  - [x] Users can add more stake to any role, `active` or `inactive`.
- Decrease Stake for Inactive Roles Only:
  - [x] Stake amounts can only be reduced for roles that currently have no active services.
  - [x] Adjustments to the stake must not fall below the minimum required amount for each role.
- Removing Roles:
  - [x] Users can remove roles only if there are no active services linked to those roles.

### Screenshot
![Image](https://github.com/webb-tools/webb-dapp/assets/109539270/7961e8c3-5b5d-4c24-bb23-9cb96de5f338)

yurixander:
I will be handling this closely tied to #1988, since the UI is essentially the same but with minor differences. The Substrate logic is also quite similar, if not nearly identical when making the transaction.

Issue #1989, opened 2024-01-23
[FEAT] Restake Shared Profile Setup Flow


monaiuu:
## Description
The shared profile option enables users to allocate a single total restake amount, which is then shared across all selected service roles. This task involves implementing the flow and logic for setting up a shared restake profile.

### Key Points:
**Profile Selection:** 
- [x] Implement functionality allowing users to select the 'Shared' profile type within the Restake UI.

**Roles/ Amount Input Selection:**
- [x] Allow users to select roles from a dropdown menu with multi-select checkboxes. 
- [x] Implement an error handling mechanism that prompts the user if the input amount is less than the minimum required for a role.
- [x] Return an error if the input amount exceeds the user's remaining stakable balance.

**Dynamic Chart Component:**
- [x] If feasible, integrate a dynamic chart component that visually represents the distribution of stakes across selected roles. Ensure that the chart updates in real-time as users adjust their stake allocations.

### Screenshot
![Image](https://github.com/webb-tools/webb-dapp/assets/109539270/ea9daeb9-d360-48fd-81ee-d73ab0b3d31d)

Related Task:

PR #2053, opened 2024-02-26
feat(tangle-dapp): Implement restaking UI: Independent profile setup

### LINKED ISSUE #1988: [FEAT] Restake Independent Profile Setup Flow
## Description
Implement functionality allowing users to select the 'Independent' profile type within the Restake UI. This flow will enable users to allocate variable amounts of their stake to each selected service independently.

### Key Points:
**Profile Selection:** 
- [x] Implement functionality allowing users to select the 'Independent' profile type within the Restake UI.

**Roles/ Amount Input Selection:**
- [x] Validate inputs to ensure they meet or exceed the minimum required amount for each role.
- [x] Allow users to select roles from a dropdown menu. Ensure that roles already selected are not displayed in the dropdown menu to prevent duplication.
- [x] Implement an error handling mechanism that prompts the user if the input amount is less than the minimum required for a role.
- [x] Disable the 'Add Role' button if the total input amount equals or exceeds the available amount for restaking.

**Dynamic Chart Component:**
- [x] If feasible, integrate a dynamic chart component that visually represents the distribution of stakes across selected roles. Ensure that the chart updates in real-time as users adjust their stake allocations.

### Screenshot
Image

yurixander:
## Summary of changes

- Implemented both UI & Substrate transaction logic for managing the independent restaking profile.
  - Not much is needed for #1991 and #1989 based on what I've implemented here, but it'll track that separately to separate concerns.
- When the user confirms their allocations, say that they don't have an existing profile, the profile is automatically created. If a profile exists, then it is updated.
- Calculated the max restaking amount from the Rust source code. It is currently defined to be `max restaking amount = 50% of staked balance`. In other words, `max restaking amount = staked balance / 2`.
  - I tried to get/calculate this from the PolkadotJS API, but it seems that it is not obtainable from there, but rather present as a compile-time constant in the Rust source code.
- Added extensive input validation, & different possible states and handling of edge cases, such as when the user has no allocations, or when the allocation amount exceeds the remaining balance available for restaking. Also added logic to safely handle & parse large input amounts, and convert them to bigints (BN), which cannot be handled by simply using `Number`s.
- Created reusable [`useRestakingAllocations`](https://github.com/webb-tools/webb-dapp/pull/2053/files#diff-bf841b954f112b69fb7a6dca74a5a804b95dbb9f4233057a54205353895017baR44), [`useRestakingRoleLedger`](https://github.com/webb-tools/webb-dapp/pull/2053/files#diff-0152b22897e0910e606ff1d6f7f11a4d255b24df62c89409bf58cc15fc60dea4R9), and [`useUpdateRestakingProfileTx`](https://github.com/webb-tools/webb-dapp/pull/2053/files#diff-443a283fde94e7327aa83d6c485211ee18ae5a9a01f096b9ddaa8f1716ee6a91R29) hooks.
  - The `useUpdateRestakingProfileTx` hook can automatically create the profile if it doesn't exist. This way it automatically handles both cases at once, keeping the logic simpler and avoiding having to repeat the same logic for profile creation & updates. Can easily be opted out from this by a parameter.
- Existing allocations for roles are loaded from the Polkadot API and displayed on the modal. The user can then edit those allocations if they wish.
- Fixed bug on Webb UI library, [made the `Tooltip` component's z-index `50`](https://github.com/webb-tools/webb-dapp/pull/2053/files#diff-2b4fde6d5b73a332c0b99d6be23c74b8739d71c30741cf2d95f058082b73c2be), so that it can be used on modals with no problem. Without this, it needs to be given `z-50` property everywhere used within a modal.
- Implemented light theme variant for the entire design I've added.
- Made the design responsive so that it looks well on mobile.
- [Should not be checking SVG files for format with Prettier](https://github.com/webb-tools/webb-dapp/pull/2053/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519).

### Other notes
- The current role list is merely a dummy list. I think @AtelyPham was working on this, need to confirm if that is still being worked on or already finished, to complete this into a truly usable state.
- There seems to be either a bug or a problem when adding more than 2 allocations. This will be tracked in a separate issue: https://github.com/webb-tools/tangle/issues/508.
- The `roles.updateProfile` and `roles.createProfile` extrinsics from the PolkadotJS API seem to have type `any`, which is strange. Maybe some type definitions are missing here? [See here](https://github.com/webb-tools/webb-dapp/pull/2053/files#diff-443a283fde94e7327aa83d6c485211ee18ae5a9a01f096b9ddaa8f1716ee6a91R60-R64).

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [x] `apps/tangle-dapp`
- [ ] `apps/faucet`
- [ ] `apps/zk-explorer`

- Closes #1988

### Screen Recording

#### Functionality

https://github.com/webb-tools/webb-dapp/assets/101931215/3712d973-b713-4110-af85-b8e10453eaf1

#### Responsiveness & Light Theme

https://github.com/webb-tools/webb-dapp/assets/101931215/0bacb1c0-6df5-4a1d-ba16-bc742468c703

---

### Code Checklist


apps/tangle-dapp/components/BnChartTooltip.tsx
```diff
@@ -0,0 +1,45 @@
+import { Typography } from '@webb-tools/webb-ui-components/typography/Typography';
+import BN from 'bn.js';
```
vutuanlinh2k2, line 2: I'm not sure why we need to use `BN` with `bn.js` lib here, `polkadotjs` already provides this

apps/tangle-dapp/components/BnChartTooltip.tsx
```diff
@@ -0,0 +1,45 @@
...
+const BnChartTooltip = (
+  allocations: Record<ServiceType, BN | null>,
+  freeBalance: BN,
+  restakedAmount: BN
+) => {
+  const composition = ({
+    active,
+    payload,
+  }: TooltipProps<number, ServiceType>) => {
+    const role = payload?.[0]?.name;
```
vutuanlinh2k2, line 17: I think we need to add `useMemo` in all variables here because the value for `active` will be updated frequently

apps/tangle-dapp/containers/ManageProfileModalContainer/AllocationInput.tsx
```diff
@@ -0,0 +1,298 @@
...
+      }
+
+      const newAmountInChainUnits = convertAmountStringToChainUnits(newValue);
+
+      onChange(newAmountInChainUnits);
+    },
+    [onChange]
+  );
+
+  const handleDelete = () => {
```
vutuanlinh2k2, line 109: I think it's best if we add `useMemo` and `useCallback` in variables and functions below

apps/tangle-dapp/containers/ManageProfileModalContainer/ChooseMethodStep.tsx
```diff
@@ -0,0 +1,110 @@
+import { Card, Typography } from '@webb-tools/webb-ui-components';
+import Image from 'next/image';
+import { useTheme } from 'next-themes';
```
vutuanlinh2k2, line 3: you can use the `useNextDarkMode` hook here

apps/tangle-dapp/containers/ManageProfileModalContainer/IndependentAllocationStep.tsx
```diff
@@ -0,0 +1,328 @@
+import { BN } from '@polkadot/util';
+import { Button, Typography } from '@webb-tools/webb-ui-components';
+import assert from 'assert';
+import { useTheme } from 'next-themes';
```
vutuanlinh2k2, line 4: you can use the `useNextDarkMode` hook here

apps/tangle-dapp/data/restaking/useRestakingAllocations.ts
```diff
@@ -0,0 +1,85 @@
+import { PalletRolesProfileRecord } from '@polkadot/types/lookup';
+import { BN } from '@polkadot/util';
+
+import { RestakingProfileType } from '../../containers/ManageProfileModalContainer/ManageProfileModalContainer';
+import { RestakingAllocationMap } from '../../containers/ManageProfileModalContainer/types';
+import { ServiceType } from '../../types';
+import useRestakingRoleLedger from './useRestakingRoleLedger';
+
+function convertRecordToAllocation(
```
vutuanlinh2k2, line 9: add comments to explain this function

apps/tangle-dapp/data/restaking/useRestakingAllocations.ts
```diff
@@ -0,0 +1,85 @@
...
+    // switch statement which would provide exhaustive static type checking.
+    throw new Error(
+      'Unknown role type (was a new role added? if so, update this function)'
+    );
+  }
+
+  return [serviceType, record.amount.unwrap()];
+}
+
+const useRestakingAllocations = (profileType: RestakingProfileType) => {
```
vutuanlinh2k2, line 46: add comments to explain this function

vutuanlinh2k2 (commented):
Left some comments

apps/tangle-dapp/components/BnChartTooltip.tsx
yurixander: Since the operations (ie. object access) are straightforward, using `useMemo` would not provide any benefit here since it is not an expensive operation that needs to be cached.

apps/tangle-dapp/containers/ManageProfileModalContainer/AllocationInput.tsx
yurixander: One of the functions I didn't add `useCallback` for, because it's not passed directly to a child as a prop, instead it needs to be wrapped in a closure `() => ...`, and I've analyzed & determined that adding `useCallback` here might not be worth the complexity added to properly accomplish that. In case that we do run into performance issues in the future, then I'll revisit it.

apps/tangle-dapp/containers/ManageProfileModalContainer/AllocationInput.tsx
```diff
@@ -0,0 +1,298 @@
...
+    }
+  }, [onDelete, service]);
+
+  const toggleDropdown = useCallback(() => {
+    if (!isDisabled) {
+      setIsDropdownVisible((isVisible) => !isVisible);
+    }
+  }, [isDisabled]);
+
+  const handleSetService = (service: ServiceType) => {
```
vutuanlinh2k2, line 121: adding `useMemo` and `useCallback` in below functions and variables (line 120 - 151)

apps/tangle-dapp/containers/ManageProfileModalContainer/IndependentAllocationStep.tsx
```diff
@@ -0,0 +1,331 @@
...
+  const { data: stakedBalance } = useStakingLedgerRx(
+    useCallback((ledger) => ledger.total.toBn(), [])
+  );
+
+  // Max restaking amount = 50% of the total staked balance, which
+  // is equivalent to dividing the staked balance by 2. This is taken
+  // from Tangle's source code, as it seems that it is not obtainable
+  // from the Polkadot API.
+  // See: https://github.com/webb-tools/tangle/blob/8be20aa02a764422e1fd0ba30bc70b99d5f66887/runtime/mainnet/src/lib.rs#L1137
+  const maxRestakingAmount = stakedBalance?.divn(2) ?? null;
```
vutuanlinh2k2, line 103: adding `useMemo` and `useCallback` in below functions and variables (from line 103 to 203)

yurixander: I've added usage of `useMemo` and `useCallback` as suggested. Do note that some of the variables/computations are simple operations and did not require `useMemo`.

vutuanlinh2k2 (approved):

yurixander:
Waiting for design review to merge, cc @monaiuu.

monaiuu (changes requested):
For this step, can we auto generate the chart distribution here after inputting the amount (without having the user select `add role`)? 

Screenshot 2024-03-07 at 1 55 55 PM

yurixander:
> For this step, can we auto generate the chart distribution here after inputting the amount (without having the user select `add role`)?
> 
> Screenshot 2024-03-07 at 1 55 55 PM

Yes. I'll handle this on my [other PR](https://github.com/webb-tools/webb-dapp/pull/2081) which is more advanced than this one, to prevent merge conflicts.

monaiuu:
> > For this step, can we auto generate the chart distribution here after inputting the amount (without having the user select `add role`)?
> > Screenshot 2024-03-07 at 1 55 55 PM
> 
> Yes. I'll handle this on my [other PR](https://github.com/webb-tools/webb-dapp/pull/2081) which is more advanced than this one, to prevent merge conflicts.

Ok sounds good!

monaiuu (approved):

PR #2211, opened 2024-04-10
Add View Account button on nomination page

vutuanlinh2k2:
## Summary of changes

- Add View Account button on nomination page

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [x] `apps/tangle-dapp`
- [ ] `apps/faucet`
- [ ] `apps/zk-explorer`

### Screenshots

CleanShot 2024-04-10 at 16 13 55@2x

CleanShot 2024-04-10 at 16 14 15@2x

---

### Code Checklist

- [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/apps/tangle-dapp/containers/NominatorStatsContainer/NominatorStatsContainer.tsx
+++ b/apps/tangle-dapp/containers/NominatorStatsContainer/NominatorStatsContainer.tsx
@@ -17,7 +17,7 @@ import { NominatorStatsItem, UnbondingStatsItem } from '../../components';
 import useNetworkStore from '../../context/useNetworkStore';
 import useIsFirstTimeNominator from '../../hooks/useIsFirstTimeNominator';
 import useNetworkFeatures from '../../hooks/useNetworkFeatures';
-import { NetworkFeature } from '../../types';
+import { NetworkFeature, PagePath } from '../../types';
 import { evmToSubstrateAddress } from '../../utils';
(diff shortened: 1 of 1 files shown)


apps/tangle-dapp/containers/NominatorStatsContainer/NominatorStatsContainer.tsx
```diff
@@ -86,6 +86,16 @@ const NominatorStatsContainer: FC = () => {
                 </Link>
               )}
 
+            <Link href={PagePath.ACCOUNT}>
+              <Button
+                variant="utility"
+                className="!min-w-[100px]"
+                isDisabled={isActiveAccountLoading}
```
yurixander, line 93: We should enable it even if there is no active account connected because the link should still function even without an account

Although I think I get why you chose to disable it, maybe because "View account" implies there's an active account?

vutuanlinh2k2: I only disable it when the account is being loaded, after that it can be null or have value but the button is still available

vutuanlinh2k2: CleanShot 2024-04-10 at 16 37 23@2x

yurixander (approved):

PR #2257, opened 2024-04-22
Have the native token symbol be hardcoded instead of dynamic to improve performance

### LINKED ISSUE #2249: [TASK] Have the token unit be hardcoded instead of dynamic
The token unit should be hardcoded instead of fetched from the API. This is because regardless of the network the users connect to, we know the token will either be `tTNT` or `TNT`.

So we can make the following assumption: If active network is mainnet, then `TNT`, otherwise `tTNT`. This is because there is only one mainnet (which corresponds to `TNT`), and all other networks (even Custom networks) would fall under testnets or development/local networks.

vutuanlinh2k2:
## Summary of changes

- Have the token unit be hardcoded instead of dynamic
- Add new Tangle constants for universal use
- Remove EVM Explorer URL for local Tangle

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [x] `apps/tangle-dapp`
- [ ] `apps/faucet`
- [ ] `apps/zk-explorer`

- Closes #2249 

---

### Code Checklist

- [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/apps/tangle-dapp/app/restake/RoleDistributionCard/DistributionChart.tsx
+++ b/apps/tangle-dapp/app/restake/RoleDistributionCard/DistributionChart.tsx
@@ -1,5 +1,6 @@
 'use client';
 
+import { TANGLE_TOKEN_DECIMALS } from '@webb-tools/dapp-config/constants/tangle';
 import assert from 'assert';
 import { FC } from 'react';
 import { formatUnits } from 'viem';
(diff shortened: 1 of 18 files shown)


drewstone (commented):
Looks good, need to fix CI

PR #2500, opened 2024-08-04
feat(tangle-dapp): Implement stake & unstake liquifier functionality for EVM-based liquid staking

### LINKED ISSUE #2446: [FEAT] Add stake & unstake liquifier functionality for Ethereum-based tokens such as Chainlink
* This will be based on @freeslugs's implementation of the liquifier, which is similar to Tenderize.me.
* Each token/chain will have a corresponding adapter which can be used/called for the various actions such as staking, unstaking, getting info about the unstaking requests, and more. The interface is standardized across all adapters, so it can be reused just with different chains.
* This will be implemented on the liquid staking UI, and will reuse the same UI. There will be an option to select `Ethereum Chainlink`, which will be a special case since it will change the active network to Ethereum so that the frontend can interact with the Solidity adapter logic. For the other cases, they are all present in the Tangle Restaking Parachain.

Relevant links:
- https://github.com/webb-tools/tnt-core/blob/main/src/lst/liquifier/Liquifier.sol
- https://github.com/webb-tools/tnt-core/blob/main/src/lst/unlocks/Unlocks.sol
- https://github.com/webb-tools/tnt-core/blob/c382a3f73164e0e30d0ad76633a905673814e823/src/lst/adapters/ChainlinkAdapter.sol

image
image

yurixander:
## Summary of changes

- ➕ Users can now liquid stake and unstake assets for EVM-based chains: Polygon, Livepeer, and Chainlink. More ERC20-based tokens can be easily added and integrated as needed.
- ➕ On development, Sepolia Testnet is used to interact with some dummy/skeleton liquifier contracts that I deployed. These contracts have the same signature as the actual contracts but with dummy data. This allows for easier testing and development of the liquifier interface.
- ➕ Added useful step indicator for the processing notification (ex. `(1/2)`), since the liquid staking process for EVM-based chains is composed of two steps (approve ERC20 for spending, then call liquifier contract).
- 🔧 Re-organized liquid staking page structure, since now it is no longer a specific page per token.
- 🔧 Generalized liquid staking logic to accommodate for non-Substrate based chains. This should make it easier to integrate new protocols in the future.
- 🐛 Fixed bug where scroll area was not scrollable when selecting/changing liquid staking networks.
- 🐛 Fixed bug where inputting large amounts into the amount input would cause an error to be thrown.

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [x] `apps/tangle-dapp`
- [ ] `apps/faucet`
- [ ] `apps/zk-explorer`

- Closes #2446

### Screen Recording

https://github.com/user-attachments/assets/ec9ace53-2105-4c7e-919a-7633db124fcc

---

### Code Checklist


drewstone (commented):
With the introduction of Liquifie, it might be better to call them `LsProtocolId` unless we will map these protocols to chains and use Ethereum chain ID as a common parameter for staking protocols on Ethereum (Chainlink, Graph, Livepeer, etc.)

yurixander:
Nearly ready to submit for review here. Just need to cleanup logic and errors.

yurixander:
In this PR I implemented a simple multi-step notification indicator (e.g. `(1/2) Processing Liquifier spending approve`), since batching doesn't seem to be supported in this case.

This process is similar to DEX swapping, where you first have to make a tx to approve spending of the ERC20 token by the swapping contract, then another to actually perform the swap.

However I had noticed a much more polished notification system in the case of restaking -- which shows the status of each step visually. Curious if it would be a good idea to abstract that into a reusable component to use for multi-step transactions? (cc @monaiuu @AtelyPham)

AtelyPham (approved):
LGTM 👍!

drewstone:
My take is that we should use chain selection for chains and token selection for protocols.
image

Right now it looks like you're conflating chain selection with protocols, which creates unnecessarily many options. Ethereum should be what you select for LINK and it should show all valid tokens for Liquid Staking on Ethereum (LINK, GRAPH, Livepeer, etc.), Polygon would be a chain as well and MATIC on that chain.

If this messes with how the naming conventions are in code we should modify things so its clear on this front. But it should be preserved that we use chains for the label that simiarly says Tangle Parachain/Polkadot/etc.

yurixander:
Ok gotcha so where it says Chainlink I would instead show the chain (Ethereum Mainnet) and then where it says LINK it would be the token on that chain? @drewstone

drewstone:
Correct @yurixander. I think that slot is best for chain selection so its clear when you're bridging from one chain to another. Therefore, it can be clear if you're liquid staking and remaining on Ethereum vs bridging to Tangle.

yurixander:
> Correct @yurixander. I think that slot is best for chain selection so its clear when you're bridging from one chain to another. Therefore, it can be clear if you're liquid staking and remaining on Ethereum vs bridging to Tangle.

Makes sense and I agree. I'll implement that in my other PR.

PR #2531, opened 2024-08-30
feat(tangle-dapp): Use `multicall3` contract to batch Liquifier withdraw calls

### LINKED ISSUE #2522: [TASK] Use `Liquifier.sol::multicall` to batch `withdraw` transactions
- Currently, when the user wants to withdraw their unlock NFTs, they have to perform a transaction PER NFT. A single transaction would be much more ideal for UX.
- The [`Liquifier.sol` contract](https://github.com/webb-tools/tnt-core/blob/21c158d6cb11e2b5f50409d377431e7cd51ff72f/src/lst/liquifier/Liquifier.sol#L24) inherits `Multicall.sol` abstract contract, and thus `multicall` can be used to batch withdraw calls into a single transaction.
- Other calls such as `deposit` and `unlock` don't need batching because they are a single operation (ie. stake & unstake).
- Alternatively, a different, public & popular `multicall` contract such as [`Multicall3` contract](https://github.com/mds1/multicall) @ `0xcA11bde05977b3631167028862bE2a173976CA11`. ⚠️ Need to be careful here since this contract isn't owned/written by us.
- One downside to using multicall would be that it obscures transaction details. So wallets like MetaMask might show a generic function like `aggregate3` (part of the multicall contract) instead of the actual target function and arguments.

yurixander:
## Summary of changes

- 🔧 Calls to `Liquifier.sol::withdraw` are now batched to enable redemption of multiple NFTs within a single transaction. Before, it had to be done via sequential transactions, which greatly affects UX for larger numbers of unlock NFTs.
- ➕ Created `useContractWriteBatch` reusable hook which utilizes [`multicall3` contract](https://github.com/mds1/multicall) to batch transactions. This hook can be used for other transactions that may require batching in the future.

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [x] `apps/tangle-dapp`
- [ ] `apps/faucet`
- [ ] `apps/zk-explorer`

- Closes #2522

### Screen Recording

Screenshot 2024-08-31 at 01 08 08

Screenshot 2024-08-31 at 07 07 11

---

### Code Checklist

### DIFF
--- a/apps/tangle-dapp/components/LiquidStaking/unstakeRequestsTable/WithdrawUnlockNftButton.tsx
+++ b/apps/tangle-dapp/components/LiquidStaking/unstakeRequestsTable/WithdrawUnlockNftButton.tsx
@@ -24,22 +24,7 @@ const WithdrawUnlockNftButton: FC<WithdrawUnlockNftButtonProps> = ({
     }
 
     setIsProcessing(true);
-
-    for (const [index, unlockId] of unlockIds.entries()) {
-      const success = await withdraw(tokenId, unlockId, {
-        current: index + 1,
-        total: unlockIds.length,
-      });
-
-      if (!success) {
-        console.error(
-          'Liquifier withdraw batch was aborted because one request failed',
-        );
-
-        break;
-      }
-    }
-
+    await withdraw(tokenId, unlockIds);
     setIsProcessing(false);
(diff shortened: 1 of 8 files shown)


AtelyPham (approved):
LGTM 👍!

PR #2544, opened 2024-09-16
feat(tangle-dapp): Add Tangle mainnet & testnet networks to liquid staking

### LINKED ISSUE #2541: [FEAT] Create `Tangle Mainnet` network entry, with its token options being LS pools
A new network option for liquid staking should be added to allow liquid staking TNT with the newly added liquid staking pools on Tangle mainnet. So we'd have a total of 3 options:
- Ethereum mainnet: Liquifier contract
- Tangle restaking parachain: Liquid staking of foreign assets (DOT, PHA, etc.)
- Tangle mainnet: TNT liquid staking

yurixander:
## Summary of changes

- ➕ Tangle networks with the `lst` pallet now appear as options for liquid staking. Before it was just the parachain and liquifier.
- ➕ Added ability to join & unstake (unbond) liquid staking pools on Tangle networks (mainnet, testnet, local). Users must choose a pool onto which to liquid stake.
- ➕ Liquid (derivative) asset balance is shown for Tangle networks, on the unstake tab. This helps the user see how much balance they have staked on a given pool.
- 🔧 Removed the feedback banner as suggested by Drew.
- 🐛 Fixed bug where URL search param sync wouldn't clear values.
- 🐛 Fixed invalid URL error thrown due to an extra argument passed to `useTxNotification` hook.

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [x] `apps/tangle-dapp`
- [ ] `apps/faucet`
- [ ] `apps/zk-explorer`

- Closes #2541 

### Screen Recording

https://github.com/user-attachments/assets/8b933b3d-1c5d-4ddd-a051-e661915f9841

---

### Code Checklist


apps/tangle-dapp/utils/formatBn.ts
```diff
@@ -29,7 +29,7 @@ const DEFAULT_FORMAT_OPTIONS: FormatOptions = {
   trimTrailingZeroes: true,
 };
 
-// TODO: Break this function down into smaller local functions for improved legibility and modularity, since its logic is getting complex. Consider making it functional instead of modifying the various variables: Return {integerPart, fractionalPart} per transformation/function, so that it can be easily chainable monad-style.
+// TODO: Break this function down into smaller local functions for improved legibility and modularity, since its logic is getting complex. Consider making it functional instead of modifying the various variables: Return {integerPart, fractionalPart} per transformation/function, so that it can be easily chainable monad-style. Also, prefer usage of Decimal.js instead of BN for better decimal handling without needing to manually handle the edge cases.
```
AtelyPham, line 32: Why not use the `formatBalance` function from PolkadotJS to address these concerns? If you're worried about technical debt, we could replicate its logic sufficiently to suit our needs, simplifying it as necessary.

AtelyPham (approved):
Left a comment, otherwise LG 👍

apps/tangle-dapp/utils/formatBn.ts
yurixander: There were some reasons why I decided to use a custom formatting function, I think it was some limitations with the `formatBalance` function, like max decimal place behavior and other things. I'll look into the exact reasons briefly and let you know

yurixander: I believe the reasons were that the `formatBalance` function didn't allow for adding commas, and also on the custom impl. there is better control over the max decimal lengths (ie. some strange behavior if the amount is really small and with the `forceUnit` parameter, where amounts that are not zero are actually returned as `0` which is misleading.)

Issue #2574, opened 2024-09-30
[BUG] Race condition in usePromise hook causes state updates on unmounted components


AtelyPham:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The [usePromise](https://github.com/webb-tools/webb-dapp/blob/develop/apps/tangle-dapp/hooks/usePromise.ts) hook is experiencing a race condition that can lead to state updates being performed on unmounted components. This occurs when the effect runs multiple times before previous promises resolve, potentially causing memory leaks and React warnings about updating state on unmounted components.

### Expected Behavior

The hook should only update state for the most recent effect execution and should not attempt to update state on unmounted components.

### Steps To Reproduce

1. Create a component using the [usePromise](https://github.com/webb-tools/webb-dapp/blob/develop/apps/tangle-dapp/hooks/usePromise.ts) hook with a rapidly changing factory function or dependencies.
2. Trigger multiple re-renders of the component in quick succession.
3. Unmount the component before all promises have resolved.
4. Observe React warnings in the console about updating state on an unmounted component.

### Anything else?

The current implementation uses an `isMounted` ref to check if the component is still mounted before updating state. However, this check is insufficient in scenarios where the effect runs multiple times before the previous promise resolves.

Proposed solution: Implement an additional `isSubscribed` flag within the effect to properly manage the subscription state for each specific effect execution. This will ensure that only the most recent effect execution can update the component's state, preventing stale updates from previous executions.

Issue #2571, opened 2024-09-27
[BUG] Incorrect Item Count on Last Page of Pagination


AtelyPham:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

On all paginated tables in the existing dApp, the number of items displayed on the last page is incorrect; it shows the configured items per page instead of the actual items from the paginated data.

### Expected Behavior

The last page should accurately display the number of items in the paginated data.

### Steps To Reproduce

1. Navigate to the `nomination` page.  
2. Scroll to the validator table.  
3. Click on the last page.  
4. Pagination consistently displays `Showing 10 out of ...`, regardless of the actual number of rows.  

This issue occurs with all tables featuring pagination.

### Screenshot

![CleanShot 2024-09-27 at 16 38 34@2x](https://github.com/user-attachments/assets/efc134fb-578b-44a2-956b-ac8789c70ed4)

PR #2575, opened 2024-09-30
feat(tangle-dapp): Integrate Blueprints APIs

AtelyPham:
## Summary of changes

- Integrate blueprint APIs.
- Resolve race condition in `usePromise` hook.
- Correct item count on the last page of `Pagination` component.

### Proposed area of change

- [ ] `apps/bridge-dapp`
- [ ] `apps/hubble-stats`
- [ ] `apps/stats-dapp`
- [x] `apps/tangle-dapp`
- [ ] `apps/faucet`
- [ ] `apps/zk-explorer`

- Closes https://github.com/webb-tools/webb-dapp/issues/2574
- Closes https://github.com/webb-tools/webb-dapp/issues/2571
- Closes https://github.com/webb-tools/webb-dapp/issues/2496
- Closes https://github.com/webb-tools/webb-dapp/issues/2495

### Screen Recording

https://github.com/user-attachments/assets/e12d581f-e248-4f5e-a44f-8d20a09dbce4


AtelyPham:
Depends on https://github.com/webb-tools/tangle/pull/771.

apps/tangle-dapp/app/blueprints/[id]/useBlueprintDetails.ts
```diff
@@ -0,0 +1,161 @@
...
+      tvlInUsd,
+      vaultTokens: delegations
+        .map(({ amount, assetId }) => {
+          const asset = assetMap[assetId];
+          if (asset === undefined) return null;
+
+          return {
+            name: asset.name,
+            symbol: asset.symbol,
+            amount: +formatUnits(amount, asset.decimals),
```
Comment (yurixander) [inline apps/tangle-dapp/app/blueprints/[id]/useBlueprintDetails.ts:153]: Should we propagate amount in `BN` type?

apps/tangle-dapp/app/restake/TableTabs.tsx
```diff
@@ -43,29 +43,29 @@ const TableTabs: FC<Props> = ({
 
   const { rewardConfig } = useRestakeRewardConfig();
 
-  // Recalculate vaults (pools) data from assetMap
+  // Recalculate vaults (vaults) data from assetMap
```
yurixander, line 46:
```suggestion
  // Recalculate vaults data from assetMap
```

apps/tangle-dapp/constants/links.ts
yurixander, line 1: We usually have the doc links on `@webb-tools/webb-ui-components`, you think it's fine if we have them locally on Tangle dapp under a consts file?

apps/tangle-dapp/data/blueprints/useBlueprintListing.ts
```diff
@@ -0,0 +1,75 @@
...
+  const { operatorTVL } = useOperatorTVL(operatorMap, assetMap);
+
+  const { result, ...rest } = useApiRx(
+    useCallback(
+      (apiRx) => {
+        if (
+          apiRx.query.services?.blueprints === undefined ||
+          apiRx.query.services?.operators === undefined
+        )
+          // TODO: Should return the error here instead of throw it
```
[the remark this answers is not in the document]
yurixander: @drewstone I'm planning to modify the `useApiRx` hook to be able to manage errors by returning them from functions like these briefly, when I do that I'll handle this TODO

apps/tangle-dapp/data/blueprints/useBlueprintListing.ts
```diff
@@ -0,0 +1,75 @@
...
+          }),
+        );
+      },
+      [operatorMap, operatorTVL, rpcEndpoint],
+    ),
+  );
+
+  return {
+    ...rest,
+    blueprints: result ?? [],
```
yurixander, line 73: Do you think we should make it explicitly `| null` instead of returning `[]`?

apps/tangle-dapp/data/blueprints/useOperatorBlueprints.ts
```diff
@@ -0,0 +1,59 @@
...
+              );
+            }),
+            catchError((error) => {
+              console.error(
+                'Error querying services with blueprints by operator:',
+                error,
+              );
+
+              return of<OperatorBlueprint[]>([]);
+            }),
```
yurixander, line 46: Ideally we'd return an error to the `useApiRx` hook here too? Might be a bit tricky since `catchError` is being used here. Maybe leave a TODO for returning error later on?

apps/tangle-dapp/data/blueprints/useOperatorBlueprints.ts
```diff
@@ -0,0 +1,59 @@
...
+          );
+      },
+      [operatorAccount],
+    ),
+  );
+
+  return {
+    blueprints: result ?? [],
```
yurixander, line 54: Do you think we should make it explicitly `| null` instead of returning `[]`?

apps/tangle-dapp/data/blueprints/utils/toPrimitiveBlueprint.ts
```diff
@@ -0,0 +1,309 @@
...
+  if (gadget.type === 'Native') {
+    return toPrimitiveNativeGadget(gadget.asNative);
+  }
+
+  if (gadget.type === 'Wasm') {
+    return toPrimitiveWasmGadget(gadget.asWasm);
+  }
+
+  throw new Error('Unknown Gadget type');
+}
```
yurixander, line 141: `switch` might be better suited here? Especially if `Container`, `Native` and `Wasm` are all the currently available possible value of `gadget.type`.

apps/tangle-dapp/data/blueprints/utils/toPrimitiveBlueprint.ts
```diff
@@ -0,0 +1,309 @@
...
+export function toPrimitiveWasmGadget({ runtime, sources }: WasmGadget) {
+  return {
+    runtime: toPrimitiveWasmRuntime(runtime),
+    sources: sources.map(toPrimitiveGadgetSource),
+  } as const;
+}
+
+export function toPrimitiveWasmRuntime(runtime: WasmRuntime) {
+  return runtime.type;
+}
```
yurixander, line 152: Might be simpler if we get rid of this function and instead access the property directly where it's used

apps/tangle-dapp/data/blueprints/utils/toPrimitiveBlueprint.ts
yurixander, line 1: In case any of these functions are **used only once**, if they are simple enough, it might be simplest to perform the action directly on the caller

apps/tangle-dapp/data/blueprints/utils/toPrimitiveBlueprint.ts
```diff
@@ -0,0 +1,309 @@
...
+      Ipfs: u8aToString(fetcher.asIpfs),
+    } as const;
+
+  if (fetcher.type === 'Testing')
+    return {
+      Testing: toPrimitiveTestingFetcher(fetcher.asTesting),
+    } as const;
+
+  throw new Error('Unknown GadgetSourceFetcher type');
+}
```
yurixander, line 194: `switch` might be more appropriate here, especially if you've covered all the possibilities of `fetcher.type`, since that would eliminate the need to throw a runtime error and instead rely on compile-type switch exhaustiveness from the type system

apps/tangle-dapp/data/blueprints/utils/toPrimitiveBlueprint.ts
```diff
@@ -0,0 +1,309 @@
...
+  } as const;
+}
+
+export function toPrimitiveArchitecture(arch: Architecture) {
+  return arch.type;
+}
+
+export function toPrimitiveOperatingSystem(os: OperatingSystem) {
+  return os.type;
+}
```
yurixander, line 254: Might be simpler if we get rid of these functions and use the properties directly on the consumers

apps/tangle-dapp/data/blueprints/utils/toPrimitiveBlueprint.ts
```diff
@@ -0,0 +1,309 @@
...
+            [
+              toPrimitiveFieldType(first),
+              toPrimitiveFieldType(second),
+            ] as const,
+        ),
+      ],
+    } as const;
+  }
+
+  return fieldType.type;
```
yurixander, line 305: `switch` might be more idiomatic here

apps/tangle-dapp/data/restake/useDelegatorTVL.ts
```diff
@@ -0,0 +1,60 @@
...
+                if (typeof assetPrice !== 'number') {
+                  return acc;
+                }
+
+                const result = safeFormatUnits(
+                  delegation.amountBonded,
+                  assetData.decimals,
+                );
+
+                if (!result.sucess) {
```
yurixander, line 34: Tiny typo here `sucess` -> `success`

apps/tangle-dapp/data/restake/useDelegatorTVL.ts
```diff
@@ -0,0 +1,60 @@
...
+                  return acc;
+                }
+
+                const formattedAmount = Number(result.value);
+
+                acc[delegation.assetId] =
+                  (acc[delegation.assetId] || 0) + formattedAmount * assetPrice;
```
yurixander, line 41: This is a bit hard to understand, should expand just a bit or brief comment? For example, explaining why it is defaulting to `0` might help

apps/tangle-dapp/data/restake/useDelegatorTVL.ts
```diff
@@ -0,0 +1,60 @@
...
+
+                return acc;
+              },
+              {} as Record<string, number>,
+            ) || {};
```
yurixander, line 46:
```suggestion
            ) ?? {};
```
`??` instead of `||` might be more appropriate here if it returns `| undefined`

apps/tangle-dapp/data/restake/useOperatorTVL.ts
```diff
@@ -0,0 +1,66 @@
...
+                  const result = safeFormatUnits(
+                    delegation.amount,
+                    asset.decimals,
+                  );
+
+                  if (!result.sucess) {
+                    return sum;
+                  }
+
+                  const formattedAmount = Number(result.value);
```
yurixander, line 32: Would `amount` be a better name? Formatted amount might imply that it's in string form. What do you think?

apps/tangle-dapp/data/restake/useRestakeDelegatorInfo.ts
```diff
@@ -56,9 +32,7 @@ export default function useRestakeDelegatorInfo() {
           }
 
           return apiRx.query.multiAssetDelegation
-            ?.delegators<
-              Option<PalletMultiAssetDelegationDelegatorDelegatorMetadata>
-            >(activeAddress ?? '')
+            ?.delegators(activeAddress ?? '')
```
yurixander, line 35: What would happen when `delegators` is called with an empty string, when the active address isn't present? Might be better to explicitly handle this edge case, in case that empty string poses issues

apps/tangle-dapp/data/restake/useRestakeOperatorMap.ts
```diff
@@ -39,37 +30,35 @@ export default function useRestakeOperatorMap(): UseRestakeOperatorMapReturnType
...
+              delegationCount: operator.delegationCount.toNumber(),
+              bondLessRequest: toPrimitiveRequest(operator.request),
+              delegations,
+              restakersCount,
+              status: toPrimitiveStatus(operator.status),
+            } satisfies OperatorMetadata;
+
+            return Object.assign(operatorsMap, {
+              [accountId.toString()]: operatorMetadataPrimitive,
+            } satisfies OperatorMap);
```
yurixander, line 56: Curious: why is `Object.assign` being used here? Perhaps a brief comment might help for future readers

apps/tangle-dapp/hooks/usePromise.ts
```diff
@@ -49,30 +49,39 @@ function usePromise<T>(factory: () => Promise<T>, fallbackValue: T) {
       return;
     }
 
+    // The `isMounted` ref checks if the component is mounted globally,
+    // while `isSubscribe` flag manages the subscription state for this specific effect.
+    // `isMounted` prevents updates after component unmount,
+    // `isSubscribe` allows proper cleanup if the effect re-runs before the promise resolves.
+    let isSubscribe = true;
```
yurixander, line 56:
```suggestion
    // `isMounted` prevents updates after component unmount,
    // `isSubscribed` allows proper cleanup if the effect re-runs before the promise resolves.
    let isSubscribed = true;
```

apps/tangle-dapp/next.config.js
```diff
@@ -15,6 +15,16 @@ const nextConfigBase = require('../../next.config.cjs');
...
   ...nextConfigBase,
 
+  images: {
+    remotePatterns: [
+      {
+        protocol: 'https',
+        hostname: '**',
+      },
+    ],
+  },
```
yurixander, line 25: Curious: what does this enable and why was it added? A comment might help here

apps/tangle-dapp/types/error.ts
yurixander, line 1: Should I start adopting this error system on my PRs as well?

apps/tangle-dapp/utils/polkadot/identity.ts
```diff
@@ -40,37 +43,54 @@ export const extractDataFromIdentityInfo = (
...
+  const api = await getApiPromise(rpcEndpoint);
+  const identityData = await api.query.identity.identityOf.multi(addresses);
+
+  return identityData.map((data) => {
+    if (data.isNone) return null;
+
+    const [registration] = data.unwrap();
+
+    return extractIdentityInfo(registration);
+  });
```
yurixander, line 95: `.flatMap` and `if (data.isNone) return []` else `... return [extractIdentityInfo(registration)]` instead? Otherwise the consumer may need to manually filter out / ignore `null` values

libs/webb-ui-components/src/components/Pagination/Pagination.tsx
```diff
@@ -38,11 +38,20 @@ export const Pagination = React.forwardRef<HTMLDivElement, PaginationProps>(
...
+      // If not the last page, return the itemsPerPage
+      if (!isLastPage) {
+        return itemsPerPage?.toLocaleString() ?? '-';
+      }
+
+      // Otherwise, calculate the remaining items on the last page
+      const remainingItems = totalItems ? totalItems % (itemsPerPage ?? 1) : 0;
+      return remainingItems > 0
+        ? remainingItems.toLocaleString()
+        : (itemsPerPage?.toLocaleString() ?? '-');
```
yurixander, line 53: Should use `EMPTY_VALUE_PLACEHOLDER` instead of `-`?

yurixander (changes requested):

apps/tangle-dapp/app/blueprints/[id]/OperatorsTable.tsx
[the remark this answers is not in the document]
Comment (AtelyPham) [reply apps/tangle-dapp/app/blueprints/[id]/OperatorsTable.tsx:1]: I can move the code directly into `page.tsx` instead of using this component.

apps/tangle-dapp/data/blueprints/useBlueprintListing.ts
AtelyPham: I set it to `[]` to simplify handling the empty state; now, I only need to check if `length === 0` instead of handling a `null` case, which only occurs during loading or errors that I've already addressed.

apps/tangle-dapp/data/blueprints/utils/toPrimitiveBlueprint.ts
AtelyPham: I delegate the conversion of the `enum` to JS primitive types to the function, allowing us to manage more complex enum types in the future without needing to create a new function.

apps/tangle-dapp/data/blueprints/useBlueprintListing.ts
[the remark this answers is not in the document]
yurixander: @AtelyPham In this specific case, since it's checking whether the pallets are available, if the feature isn't supported we should probably have an error state or at least default to 'empty,' since this is a possible edge case that the user would run into. For example, simply switching to mainnet (where the pallet(s) might not be live yet) would trigger this error to be thrown

apps/tangle-dapp/data/blueprints/utils/toPrimitiveBlueprint.ts
AtelyPham: I want to separate the conversion logic for each type into its own function to enhance readability and maintain separation of concerns. Additionally, since these types are utilized elsewhere in the backend, I will keep these functions separate.

apps/tangle-dapp/data/restake/useRestakeDelegatorInfo.ts
AtelyPham: The empty string returns `Option.None`, but I agree that I need to handle the case when `activeAddress === null`.

apps/tangle-dapp/utils/polkadot/identity.ts
AtelyPham: I need to add documentation for this function. Its purpose is to return an array of `IdentityType` or `null`, where each element corresponds to an address in the input array. If an address lacks identity information, the corresponding element will be null.

libs/webb-ui-components/src/components/Pagination/Pagination.tsx
AtelyPham: `EMPTY_VALUE_PLACEHOLDER` is part of the `tangle-dapp`, but this component belongs to `webb-ui-components`, making the constant unable to import here.

apps/tangle-dapp/data/blueprints/useBlueprintListing.ts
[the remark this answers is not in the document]
AtelyPham: When this error occurs, the UI handles it by displaying a message to inform users that the feature is not supported.

PR #2653, opened 2024-11-13
feat(tangle-dapp): Show LST icons from pool metadata, if set

### LINKED ISSUE #2640: [FEAT] Display LS pools icon from metadata (if set)
The `icon` metadata field was recently added for liquid staking pools.  Need to adjust the logic to display the icon on the liquid staking pool tables (if it is set on the metadata) and also on the `Select Token` modal.

### LINKED ISSUE #2655: [BUG] Large payout batch causes tx to fail with block limit exceeded
Using the `Payout All` button for a large amount of payout transactions causes the transaction to fail due to exceeding the weight limit.

- [x] Instead of calculating the exact amount of payout tx's based on their weight and the max weight limit, use an optimistic payout max batch size of 40, since there's always the possibility that other txs included in the same block can push the weight and exceed the limit anyways.
- [x] In order to "guarantee" success, a priority tip would be needed, but using a max batch size of 40 works most of the time, the only downside being that the user needs to call the `Payout All` button multiple times.
- [x] This requirement should be documented on the modal.

yurixander:
## Summary of changes

- ➕ LST icons are now loaded if they're present in LS pools metadata and shown everywhere the LST's icon appears.
- ➕ Created `formatDisplayAmount` utility function to standardize balance formatting for UIs.
- 🔧 Adjusted the precision of some of the balances shown across tables, such as reducing precision for TVLs to reduce the space taken in the case of very small amounts.
- 🐛 Added missing `decimals` prop to some token amount table cells that were missing it, which was causing precision problems.
- 🐛 Fixed a bug when claiming a lot of payouts, now it limits it to 40 to avoid exceeing the block weight limit. This means that the `Payout All` button will have to be used multiple times if there are more than 40 pending payouts.

### Proposed area of change

- [x] `apps/tangle-dapp`
- [ ] `apps/tangle-cloud`
- [ ] `libs/tangle-shared-ui`

### Associated issue(s)

- Closes #2640
- Closes #2655

### Screen Recording

Screenshot 2024-11-13 at 23 17 49

Screenshot 2024-11-15 at 01 23 18


yurixander:
@AtelyPham Getting an error after updating Substrate types package:
Screenshot 2024-11-25 at 22 15 20

Also, I made _**two updates**_ to the `apps/tangle-dapp/data/restake/RestakeTx/substrate.ts` file since the update of the types added a new parameter. Let me know if those are the right parameters 👀 .
Screenshot 2024-11-25 at 22 16 13

PR #2675, opened 2024-11-26
feat(tangle-dapp): Implement LS pools update roles modal

### LINKED ISSUE #2654: [FEAT] LS pool modal: Update roles
Add a modal that allows users with the `root` role of an LS pool to update the roles of the pool. This modal should be simple, with three inputs for the different possible roles.

### LINKED ISSUE #2692: [TASK] Create & use branded address types
Currently, all different kinds of addresses are under the `string` type. This has lead to numerous logic bugs because certain functions accept different kinds of addresses and so on. By having three types of addresses (that all resolve to `string`), it'll make function signatures much stronger in terms of type safety and help reduce future logic bugs.

The different address types:
- `SubstrateAddress`
- `EvmAddress`
- `Substrate32BytesAddress` (mainly used for precompiles)

### LINKED ISSUE #2693: [TASK] Improve amount/balance formatting consistency
Currently, different places of the dApp use different formatting functions for their balances/amounts. Because of the various config options that the format function accepts, the consistency isn't great and in some cases not ideal at all (ex. depending on the formatting config., small balances show up as 0 instead of something more reasonable like `<0.0001` but not exactly 0, which would be misleading).

- [x] Implement the `formatDisplayAmount` with 3 standard formatting variants.
- [x] Use the `formatDisplayAmount` throughout the dApp as much as possible to improve consistency.

yurixander:
## Summary of changes

- ➕ Added the ability to update LS pool roles via a modal that initiates a transaction. This option automatically comes up if the user has the `root` role of any pools.
- 🔧 Improved amount/balance formatting consistency across the dApp.
- 🔧 Refactored code to make consistent use of the different address types: `SubstrateAddress`, `Substrate32ByteAddress`, and `EvmAddress`. By differentiating each type, it helps prevent possible logic bugs in the future and also improves DX.
- 🐛 Fixed bug with the Update Commission modal where the modal footer was duplicated.
- 🐛 Fixed the LST create pool precompile call since it was updated with new parameters on Tangle.

### Proposed area of change

- [x] `apps/tangle-dapp`
- [ ] `apps/tangle-cloud`
- [ ] `libs/tangle-shared-ui`

### Associated issue(s)

- Closes #2654
- Closes #2692
- Closes #2693

### Screen Recording

https://github.com/user-attachments/assets/57d8924d-1e6d-41ed-ae74-c614dc346880


yurixander:
@AtelyPham Something recently seems to have changed that was causing the build/checks to fail, first something about WASM (which I've fixed) but also about Storybook build failing. I've removed the Storybook workflow temporarily as a workaround to be able to merge this PR.

Speaking of, do you think we should get rid of Storybook? I haven't been using it for a long time, unsure if it is actively providing many benefits tbh. Thoughts?

drewstone (approved):

drewstone:
Demo looks great!

PR #2730, opened 2025-01-09
feat(tangle-cloud): Add Operator Registration Flow

AtelyPham:
## Summary of changes

_Provide a detailed description of proposed changes._

- Implement `select` and `form` components in the `webb-ui` library.
- Revise `modal` components to function without state dependency.
- Remove the `headlessui` library.
- Introduce checkboxes for operator registration in the blueprint marketplace.
- Create a register panel (with animation) on the blueprint marketplace page.
- Add pricing modals (global and individual) with form validation.
- Develop a review page to assess selected blueprints and configure registration parameters.
- Integrate logic for pre-registering and registering blueprints for operators.

### Proposed area of change

_Put an `x` in the boxes that apply._

- [x] `apps/tangle-dapp`
- [x] `apps/tangle-cloud`
- [x] `libs/tangle-shared-ui`
- [x] `libs/webb-ui-components`

### Associated issue(s)

_Specify any issues that can be closed from these changes (e.g. `Closes #233`)._

- Closes https://github.com/tangle-network/dapp/issues/2553
- Closes https://github.com/tangle-network/dapp/issues/2554
- Closes https://github.com/tangle-network/dapp/issues/2557

### Screen Recording

_If possible provide screenshots and/or a screen recording of proposed change._

https://share.cleanshot.com/t2LhZlR4

PR #2732, opened 2025-01-10
fix(tangle-dapp): Fix restake tables & refactor restaking hooks

### LINKED ISSUE #2733: [TASK] Fix Restaking hooks after Tangle node updates
Some restaking-related pallets were shuffled around by recent Tangle node changes. Need to update the restaking hooks to match these changes. Specifically, the reward config query was affected.

- [x] Update the Tangle Substrate types package
- [x] Fix and refactor any broken hook logic due to pallet changes

### LINKED ISSUE #2734: [BUG] Unstake & withdraw restaking tables only show one row
Fix a bug where the restaking unstake & withdraw tables show at most one row, even when there are more requests.

- [x] Fix bug in the unstake table.
- [x] Fix bug in the withdraw table.

yurixander:
## Summary of changes

- 🔧 Refactored & improved some of the restaking hooks.
- 🐛 Fixed a bug in the unstake & withdraw restaking tables, where a single row was shown even when there were more (#2734).
- 🐛 Fixed some restaking hooks that were outdated due to recent changes on the Tangle node (#2733).
- 🐛 Fixed a bug that was causing some modals to not open, caused by #2730.

### Proposed area of change

- [x] `apps/tangle-dapp`
- [ ] `apps/tangle-cloud`
- [ ] `libs/tangle-shared-ui`

### Associated issue(s)

- Closes #2733
- Closes #2734

### Screen Recording


yurixander:
The reason why the assets are not showing up on the testnet is because they don't have any associated restaking vault yet @monaiuu

apps/tangle-dapp/src/data/liquidStaking/useAssetAccounts.ts
```diff
@@ -9,7 +9,7 @@ import { map } from 'rxjs';
 import useNetworkFeatures from '../../hooks/useNetworkFeatures';
 import { NetworkFeature } from '../../types';
 
-const useLsPoolMembers = ():
```
AtelyPham, line 12: For the _// TODO: Manually casting the type here, since the type is being inferred as `Codec` instead of `Option<PalletAssetsAssetAccount>`._ Sometimes it happens when you use `.entries()`, you can help TS augment the type by passing the result type to the generic `entries<ActualTypeHere>()`.

AtelyPham, line 12: Do you think this hook could return massive data (and some of them maybe we don't really need), because everyone in our system can create an asset, that leads us to the case when we have a bundle of assets and a bundle of users but some of them are actually what we care about.

libs/tangle-shared-ui/src/components/ListModal.tsx
```diff
@@ -85,9 +85,8 @@ const ListModal = <T,>({
   const isEmpty = !isSearching && !isLoading && processedItems.length === 0;
 
   return (
-    <Modal>
```
AtelyPham, line 88: Don't forget to pass the `onOpenChange` prop here 😄

libs/tangle-shared-ui/src/components/ListModal.tsx
```diff
@@ -85,9 +85,8 @@ const ListModal = <T,>({
...
+    <Modal open={isOpen}>
       <ModalContent
-        isOpen={isOpen}
         onInteractOutside={() => setIsOpen(false)}
```
AtelyPham, line 90: We don't need this if we pass the `onOpenChange` prop.

AtelyPham (commented):
Left some comments, sorry I think I made a mistake on my PR when merging so that all the modals messed up. I will check again to see if there are any modals missing. Thank you.

AtelyPham:
Submit a small PR to update all missing modal states: https://github.com/tangle-network/dapp/pull/2739.

apps/tangle-dapp/src/data/liquidStaking/useAssetAccounts.ts
yurixander: I think for now this is OK, since RX queries can handle a lot of data without slowing it down compared to making requests, and also would need to define a criteria to filter out user-created assets and leave only those that are of interest

PR #2740, opened 2025-01-15
feat(tangle-dapp): Add ability for EVM accounts to use restaking operations

### LINKED ISSUE #2738: Add EVM account restaking support

yurixander:
## Summary of changes

- ➕ Significantly improved the type safety of the `useAgnosticTx` hook, now it also type checks for correct EVM precompile argument types depending on what function is used. This means that it is now much more strongly typed and will throw a compile-time error when the ABI is updated, reducing future logic bugs.
- ➕ Added initial price fetching logic, not much left to complete it but I'll finish it up on a separate PR.
- 🔧 Added missing parameters for the restaking precompile functions (deposit, delegate, etc.), since new parameters were added on recent Tangle node changes. This was causing EVM restaking transactions to fail.
- 🔧 Did some heavy refactoring & organization, moved some files to some more appropriate locations and cut down on folders. This is the reason you may see more files changed than usual (~3,000 redundant LOC removed).
- 🔧 Refactored restaking class-based API into hooks, to maintain consistency.
- 🐛 Fixed a bug in the `useAgnosticTx` hook where a success message was shown even when the transaction failed.

### Proposed area of change

- [x] `apps/tangle-dapp`
- [ ] `apps/tangle-cloud`
- [ ] `libs/tangle-shared-ui`

### Associated issue(s)

- Closes #2738

### Screen Recording

https://github.com/user-attachments/assets/0f419ed7-feec-4c7d-b77d-43263987dcdb


AtelyPham:
I appreciate the work done on this PR. I wanted to share some thoughts on the design for the restaking feature. I believe we initially agreed on maintaining a class design for this feature.

If parameters change at runtime, it seems we would only need to modify the methods within the restaking class, which could simplify maintenance. With the current approach of using React hooks for each transaction, it could lead to a large number of files if our app scales, as many of them would share similar code with only slight variations. This might complicate updates if changes are needed at runtime.

AtelyPham:
Additionally, the current design feels quite specific to the Tangle dApp, making it challenging to implement in Tangle Cloud without significant refactoring. A class design could provide a more generalized solution, keeping related functionality in just a couple of files.

I’d love to hear your thoughts on this and any plans for reusing these hooks in Tangle Cloud. I’m reviewing this PR right now!

Thanks

yurixander:
- [ ] Revert restaking hooks into classes: Substrate
- [ ] Revert restaking hooks into classes: EVM

yurixander:
Merging this one @AtelyPham if you find some suggested changes LMK I'll address them on my next PR, just getting this one in so that I can move on to the other tasks that depend on this logic

PR #2742, opened 2025-01-15
feat(tangle-dapp): Add `Join Operator` button & modal

### LINKED ISSUE #2741: [FEAT] Add "Join operators" button & modal for Substrate accounts
Create a banner on the Restaking page with the Join Operators button on it to allow Substrate accounts to register as operators.

- [x] Implement the banner with the action button.
- [x] After clicking the button, a modal with the bond amount should pop up.
- [x] Fetch and display min. bond
- [ ] ~~Fetch and display unbond conditions~~
- [x] Implement the transaction.

yurixander:
## Summary of changes

- ➕ Added the `Join Operator` banner, button and modal, as well as the transaction logic.
- ➕ Added a prop for the `Button` component which allows a tooltip to be displayed when it is disabled.

### Proposed area of change

- [x] `apps/tangle-dapp`
- [ ] `apps/tangle-cloud`
- [ ] `libs/tangle-shared-ui`

### Associated issue(s)

- Closes #2741

### Screen Recording

banner & button:
Screenshot 2025-01-15 at 15 19 15
modal:
Screenshot 2025-01-15 at 15 19 46
when an evm account is connected:
Screenshot 2025-01-15 at 15 27 49
mobile viewport:
Screenshot 2025-01-15 at 16 01 57
mobile viewport modal:
Screenshot 2025-01-15 at 16 02 13

### DIFF
--- a/libs/webb-ui-components/src/components/IconWithTooltip/IconWithTooltip.tsx
+++ b/libs/webb-ui-components/src/components/IconWithTooltip/IconWithTooltip.tsx
@@ -1,7 +1,6 @@
 import { type FC } from 'react';
 import { twMerge } from 'tailwind-merge';
 
-import { Typography } from '../../typography/Typography';
 import { Tooltip, TooltipBody, TooltipTrigger } from '../Tooltip/Tooltip';
 import { IconWithTooltipProp } from './types';
 
@@ -24,11 +23,7 @@ const IconWithTooltip: FC<IconWithTooltipProp> = ({
         {icon}
       </TooltipTrigger>
 
-      <TooltipBody {...overrideTooltipBodyProps}>
-        <Typography variant="body2" className="text-center break-normal">
-          {content}
-        </Typography>
-      </TooltipBody>
+      <TooltipBody {...overrideTooltipBodyProps}>{content}</TooltipBody>
     </Tooltip>
   );
 };
--- a/libs/webb-ui-components/src/components/buttons/Button.tsx
+++ b/libs/webb-ui-components/src/components/buttons/Button.tsx
@@ -8,6 +8,7 @@ import { ButtonContentProps, ButtonProps } from './types';
 import { useButtonProps } from './use-button-props';
 import { getButtonClassNameByVariant } from './utils';
 import { forwardRef } from 'react';
+import { Tooltip, TooltipBody, TooltipTrigger } from '../Tooltip';
 
 /**
  * The Webb Button Component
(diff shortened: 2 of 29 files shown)


yurixander:
The most up-to-date text/copy for the banner is:

```
With Tangle, operators and their delegators get rewarded by hosting blueprints. Interested in becoming an operator? Bond some tokens to get started.
```

Let me know if you have something else in mind or if it's fine @drewstone.

libs/webb-ui-components/src/components/IconWithTooltip/IconWithTooltip.tsx
```diff
@@ -24,11 +23,7 @@ const IconWithTooltip: FC<IconWithTooltipProp> = ({
         {icon}
       </TooltipTrigger>
 
-      <TooltipBody {...overrideTooltipBodyProps}>
-        <Typography variant="body2" className="text-center break-normal">
-          {content}
-        </Typography>
-      </TooltipBody>
+      <TooltipBody {...overrideTooltipBodyProps}>{content}</TooltipBody>
```
AtelyPham, line 26: This change would break the UI in other places that used this component; have you checked all other places? Instead, I suggest adding a check here. If the content is a primitive type (we have a function called `isPrimitive` in the `dapp-types` lib), render it on the `Typography`; otherwise, use the content directly without the typo component.

libs/webb-ui-components/src/components/buttons/Button.tsx
```diff
@@ -94,6 +96,18 @@ const Button = forwardRef<HTMLElement, ButtonProps>((props, ref) => {
       )}
     </Component>
   );
+
+  if (disabledTooltip === undefined || !isDisabled) {
```
AtelyPham, line 100: `disabledTooltip !== undefined && isDisabled`

We should keep the default behavior if the button is disabled but not render it with the tooltip.

AtelyPham (commented):
Left some comments.

libs/webb-ui-components/src/components/IconWithTooltip/IconWithTooltip.tsx
yurixander: @AtelyPham What I've done to compensate for removing Typography there was [give Tooltip component itself body2 and break-normal, font-normal classes](libs/webb-ui-components/src/components/Tooltip/Tooltip.tsx). If we check whether it's primitive, it may also break things since before, everything was wrapped in a Typography component... I'll look into it.

yurixander: Just checked all of its usages, everything should be fine 👍🏼

PR #2758, opened 2025-01-26
feat(tangle-dapp): Use TX relayer for EVM accounts with no balance

### LINKED ISSUE #2755: [FEAT] Add Transaction Relayer Support for EVM Users on dApp
## Description 
Currently, EVM users need TNT to pay gas fees for transactions on Tangle Network. This creates friction for new users coming from EVM chains. Integrate the transaction relayer service to handle transactions for EVM users with 0 TNT balance.

The system will check if users have zero TNT and are on EVM. If both conditions are met, transactions will be signed and sent to the TX relayer instead of submitting directly onchain. Users with TNT balance will continue with the normal transaction flow.

### Implementation
- Add wallet balance check functionality
- Implement EVM chain detection
- Modify transaction submission flow
- Integrate with transaction relayer service

### References
https://github.com/tangle-network/txrelayer-blueprint
https://github.com/tangle-network/txrelayer-blueprint/blob/main/API.md
https://github.com/tangle-network/txrelayer-blueprint/blob/main/tests/README.md

yurixander:
## Summary of changes

- ➕ Added the ability for EVM accounts to use TX relayer, which relays transactions for free.
- 🔧 Removed legacy Restaking Parachain code & hooks; LST pallet replaced most functionality.
- 🐛 Fixed LST creation flow which was failing due to updates on the ABI.
- 🐛 Fixed some bugs with the `useContractRead` and `usePolling` hooks.

### Proposed area of change

- [x] `apps/tangle-dapp`
- [ ] `apps/tangle-cloud`
- [ ] `libs/tangle-shared-ui`

### Associated issue(s)

- Closes #2755

### Screen Recording

https://github.com/user-attachments/assets/1caac061-505b-4a26-80e4-f378df9bdd77


yurixander:
What's left here:
- [x] Integrate into the `useEvmPrecompileCall` hook, ensuring that all precompile calls automatically integrate this feature.
- [x] Estimate contract call gas to provide to relayer as `gaslimit` param.
- [x] Check if the active account is an EVM account and has no balance before attempting to relay TX.
- [x] Check whether the extrinsic that the EVM account is using is elegible for TX relayer subsidy (not all are).
- [ ] Fix CORS error which occurs when making requests from localhost to testnet relayer (CC @shekohex)
- [ ] Fix `VM Exception while processing transaction: revert Invalid permit` that might be caused by mismatching chain ID on the Tangle node's code/logic (CC @shekohex)
- [ ] Optional: Instead of just checking if the user's balance is 0, check if their balance is lower than that required to execute the TX's estimated gas. This would allow accounts that have low TNT balances to also benefit from the TX relayer.

apps/tangle-dapp/src/hooks/useEvmGasEstimate.ts
AtelyPham, line 1: We have the hook `apps/tangle-dapp/src/hooks/useEvmPrecompileFee.ts` for fetching fees and gas; it may need refactoring or removal if it's unnecessary.

apps/tangle-dapp/src/hooks/useEvmCallPermitNonce.ts
```diff
@@ -0,0 +1,30 @@
...
+import useAgnosticAccountInfo from '@webb-tools/tangle-shared-ui/hooks/useAgnosticAccountInfo';
+import useContractRead from '../data/evm/useContractRead';
+import { PrecompileAddress } from '../constants/evmPrecompiles';
+import { assertEvmAddress } from '@webb-tools/webb-ui-components';
+import { useCallback } from 'react';
+
+const useEvmCallPermitNonce = (): bigint | null => {
+  const { evmAddress } = useAgnosticAccountInfo();
+
+  const { value: nonce } = useContractRead(
```
AtelyPham, line 11: Why not use `useContractRead` from `wagmi` instead of implementing a custom version? It could simplify our code and reduce maintenance efforts. I noticed you added logic for metrics, but since you're still using readContract from `viem`, it seems like it might add complexity.

apps/tangle-dapp/src/hooks/useEvmTxRelayer.ts
```diff
@@ -0,0 +1,340 @@
...
+    async <
+      Abi extends AbiFunction[],
+      FunctionName extends ExtractAbiFunctionNames<Abi>,
+    >(
+      abi: Abi,
+      precompileAddress: PrecompileAddress,
+      functionName: FunctionName,
+      args: FindAbiArgsOf<Abi, FunctionName>,
+    ): Promise<Error | EvmTxRelaySuccessResult> => {
+      assert(isReady);
```
AtelyPham, line 259: To prevent UI crashes, consider wrapping this assertion in a try-catch block.

apps/tangle-dapp/src/hooks/useEvmPrecompileCall.ts
```diff
@@ -99,10 +109,40 @@ function useEvmPrecompileAbiCall<
...
+        isEvmTxRelayerEligible(precompileAddress, factoryResult.functionName);
+
+      // Relay the transaction if the EVM account doesn't have enough to
+      // cover the transaction fees. This is like a subsidy for EVM accounts
+      // without TNT. Not all precompile functions are eligible for this; only
+      // those that are whitelisted by the EVM transaction relayer.
+      if (isEligibleForEvmTxRelayer) {
+        console.debug('Attempting to relay transaction for EVM account.');
+
+        const result = await relayEvmTx(
```
AtelyPham, line 123: This function's assertion may crash the UI; consider placing the call within a try-catch block or moving the assertion in `relayEvmTx` to a try-catch to ensure proper error handling.

apps/tangle-dapp/src/hooks/useViemWalletClient.ts
AtelyPham, line 1: Why don't we use [`useWalletClient`](https://wagmi.sh/react/api/hooks/useWalletClient#usewalletclient) from and [`usePublicClient`](https://wagmi.sh/react/api/hooks/usePublicClient) from `wagmi` out of curiosity?

AtelyPham (commented):
Left some comments.

libs/tangle-shared-ui/src/context/useProviderStore.ts
AtelyPham, line 1: We also want to [persist](https://zustand.docs.pmnd.rs/integrations/persisting-store-data) the store to local storage, so that users do not need to reconnect again after refreshing the dApp.

apps/tangle-dapp/src/hooks/useEvmTxRelayer.ts
yurixander: Has it actually crashed the UI for you? 👀 Assertions here aren't meant as an error handling mechanism, but rather logical assumptions that are always meant to hold true, mainly used to overcome some limitations that TypeScript's type system can't handle, or at least not without much complexity.

So there shouldn't be any error handling for the logical assertions, but if it is crashing the UI then yeah I could make it do something so that it doesn't crash it, just reports the error.

apps/tangle-dapp/src/hooks/useEvmPrecompileCall.ts
yurixander: Has it actually crashed the UI for you? 👀 Assertions here aren't meant as an error handling mechanism, but rather logical assumptions that are always meant to hold true, mainly used to overcome some limitations that TypeScript's type system can't handle, or at least not without much complexity.

So there shouldn't be any error handling for the logical assertions, but if it is crashing the UI then yeah I could make it do something so that it doesn't crash it, just reports the error.

apps/tangle-dapp/src/hooks/useEvmCallPermitNonce.ts
yurixander: I did consider this before creating the `useContractRead` functions some time ago, but there were some limitations with it, for example if I wanted to wait until I had some data available before making the read requests (ex. active account address was still loading or unset), and some other things of that nature. Another example would be the `useContractWrite` hook, which has the added benefit of automatic tx status reporting, which we would otherwise need to handle inside a big `useEffect` manually. This is especially important for when I unify the tx status' UI into one drawer, which would simplify it: I would just need to configure `useContractWrite` instead of all the places where we would be using wagmi hooks.

TLDR: Better integration with our dapp's needs, more flexibility, and auto tx status reporting.

apps/tangle-dapp/src/hooks/useEvmGasEstimate.ts
yurixander: I noticed that; I'll be fixing/removing/replacing it later on though, as to not break existing stuff dependent on it, and also trying to avoid expanding the scope of this PR as much as possible.

apps/tangle-dapp/src/hooks/useViemWalletClient.ts
yurixander: This is integrated with `useNetworkStore`'s active network, so that when the active network is changed, this also reflects the changes. Is it the same for `useWalletClient` & `usePublicClient`?

libs/tangle-shared-ui/src/context/useProviderStore.ts
yurixander: Agreed; this one's more on the 'building' state though, just laying out foundations for it.

apps/tangle-dapp/src/hooks/useEvmTxRelayer.ts
AtelyPham: Assertions are meant to confirm a condition is always true, but you only check the `isReady` variable without additional use. Instead, consider using `if (!isReady) return Error()`, which is more straightforward and safer, particularly since the function can return an error.

In my view, every time I see an assertion, I assume it can throw an error.

apps/tangle-dapp/src/hooks/useViemWalletClient.ts
AtelyPham: Yes, you can pass the [chain id](https://wagmi.sh/react/api/hooks/useWalletClient#chainid) as the parameter into those hooks.

apps/tangle-dapp/src/hooks/useEvmTxRelayer.ts
yurixander: > Assertions are meant to confirm a condition is always true, but you only check the isReady variable without additional use.

By asserting `isReady`, the TypeScript compiler is smart enough to know that all the variables used within the `relayEvmTx` function aren't undefined or null. Check out what happens when I remove that assertion:
Screenshot 2025-01-30 at 17 56 33

This condition should ensure that that error never occurs, and if it does, it means that the code has been modified in an unexpected way, in which case, _it should throw the error_, as it'll let the developer know that there's a logic bug at runtime. This is the reasoning for why the assertion should never fail BTW:
Screenshot 2025-01-30 at 17 57 49

> consider using if (!isReady) return Error(), which is more straightforward and safer, particularly since the function can return an error.

In this case, I think it is perfectly fine to leave it as an assertion instead of a handled error, since it is intended to be loud and clear if the assertion isn't met rather than being subtle, because it would mean that there's a logic bug introduced by the developer and not an error that would occur under normal circumstances when calling `relayEvmTx`.

CC @AtelyPham

apps/tangle-dapp/src/hooks/useViemWalletClient.ts
yurixander: How would you obtain that chain id? Something like this would be needed wherever we use `useWalletClient` & `usePublicClient`:

```tsx
const MyComponent = () => {
  const {network} = useNetworkStore()
  
  // ... check if the current network has an evm id ...
  
  const client = useWalletClient({ chainId: network.evmChainId, ... other config ... })
}
```

If that's the case, then _that would be exactly why_ the custom `useViemWalletClient` hook is useful: It saves you the time to having to repeat all those checks wherever you'd use the `wagmi` variant.

drewstone:
@yurixander after depositing does your balance instantly update or not in that tooltip display?

drewstone:
Cause if not we're just leaving more "BUG"s that @monaiuu is going to have us point out. Remember these features don't exist in a vacuum we have to be implementing updates for all the effects they trigger.

yurixander:
> @yurixander after depositing does your balance instantly update or not in that tooltip display?

@drewstone It does not instantly update after depositing, the reason likely being that the code fetches the balances periodically. It is not fixed in this PR because this one is solely intended to implement the TX relayer feature though.

drewstone:
Ok, is this bug reported anywhere?

drewstone:
I think if we don't change this culture of how we build where we're so focused on implementing a feature and not doing it that fixes things in the process we will continue to have issues like this.

yurixander:
> Ok, is this bug reported anywhere?

Yes: #2780

PR #2786, opened 2025-01-31
chore(tangle-dapp): Add token to wallet post bridge transaction

### LINKED ISSUE #2767: [TASK] Implement 'Add Token to Wallet' Post-Bridge Transaction
## Description
After making a successful bridge transaction, we should prompt user to add the associated to their wallet so that user can see their bridged tokens on Tangle immediately.  This should also allow user to see the bridged token within the restake vaults table. 

### Screenshot
Image

devpavan04:
## Summary of changes

- After executing a bridge transaction, the user is prompted to add the bridged token to their wallet.

### Proposed area of change

- [x] `apps/tangle-dapp`
- [ ] `apps/tangle-cloud`
- [x] `libs/tangle-shared-ui`

### Associated issue(s)

- Closes #2767 

### Screen Recording

https://github.com/user-attachments/assets/f72178b1-966f-4a61-941e-8613f4f4eb22


monaiuu (approved):

yurixander (commented):
@devpavan04 Does this PR handle the case where the token already exists in the wallet? Will it be prompted to re-add it?

devpavan04:
> @devpavan04 Does this PR handle the case where the token already exists in the wallet? Will it be prompted to re-add it?

Currently, there is no direct method to check if a token is already added. The MetaMask API does not provide a function to retrieve a list of tokens added to a user's wallet.

It will be prompted to re-add it, but the second modal will have a warning - already exists, be careful to add unknown tokens, etc.

@monaiuu We can go with the `add to wallet` button approach here. What do you think? But still, we cannot hide/show the button since we do not know if the token has already been added.

AtelyPham (approved):

AtelyPham:
> > @devpavan04 Does this PR handle the case where the token already exists in the wallet? Will it be prompted to re-add it?
> 
> Currently, there is no direct method to check if a token is already added. The MetaMask API does not provide a function to retrieve a list of tokens added to a user's wallet.
> 
> It will be prompted to re-add it, but the second modal will have a warning - already exists, be careful to add unknown tokens, etc.
> 
> @monaiuu We can go with the `add to wallet` button approach here. What do you think? But still, we cannot hide/show the button since we do not know if the token has already been added.

We can use local storage to track if a token has been added to the MetaMask wallet. If the user successfully adds the token or receives the "Token Already Exists" error, we can record that the token is added for their currently selected account. However, this approach may require additional implementation.

Alternatively, we could display the "Add to Wallet" button. If the token is already in the user's wallet, clicking the button would simply inform them that the token exists.

devpavan04:
Checklist:

- [ ] Use local storage to store token <> wallet info.

PR #2846, opened 2025-02-25
fix(tangle-shared-ui): Fix Talisman Keep Popping Up

AtelyPham:
## Summary of changes

- Fix the issue of the Talisman wallet repeatedly appearing during page re-renders.
- Ensure that only the selected wallet opens when connecting to a Substrate wallet, rather than all available Substrate wallets.
- Call `web3Enable` only once, and use `web3EnablePromise` if it is initialized. 
	- **Note:** _Ideally, avoid calling `web3Enable` as it triggers the connection for all installed Substrate wallets._
- Utilize the injected extension from the active API instead of reinitializing and searching for the target extension, which also causes all unconnected Substrate wallets to appear.
- Simplify the `sign` method by directly using the injected extension signer from the API.

### Proposed area of change

- [x] `apps/tangle-dapp`
- [ ] `apps/tangle-cloud`
- [x] `libs/tangle-shared-ui`
- [ ] `libs/ui-components`

### Associated issue(s)

- Resolved 2 items in https://github.com/tangle-network/dapp/issues/2835

### Screen Recording

https://github.com/user-attachments/assets/d38da6db-cdd4-4f5c-9d3b-398c854838af

### DIFF
--- a/libs/dapp-config/src/types.ts
+++ b/libs/dapp-config/src/types.ts
@@ -1,4 +1,10 @@
-// File contains shared types for dapp-config
+import { InjectedWindowProvider } from '@polkadot/extension-inject/types';
+
+declare global {
+  interface Window {
+    injectedWeb3?: Record<string, InjectedWindowProvider>;
+  }
+}
 
 /**
  * The type of the environment the dapp is running in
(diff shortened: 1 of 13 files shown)


AtelyPham:
We attempt to connect to the RPC during API initialization, which causes the delay.

yurixander (approved):

PR #2885, opened 2025-03-09
feat(tangle-dapp): Add Phantom wallet to dApp wallet provider

### LINKED ISSUE #2809: [TASK] Add Phantom wallet to dApp wallet provider

devpavan04:
## Summary of changes

- add `solana-api-provider`.
- add `Phantom` wallet to dApp.

### Proposed area of change

- [x] `apps/tangle-dapp`
- [ ] `apps/tangle-cloud`
- [x] `libs/tangle-shared-ui`
- [x] `libs/ui-components`

### Associated issue(s)

- Closes #2809 

### Screen Recording & Screenshots

CleanShot 2025-03-09 at 12 05 18@2x


drewstone:
CI needs fix, what flows have you QAd with this?

yuri-xyz:
By the way, might be best to leave new wallet/provider integrations post-restaking launch, since the system will likely get re-worked so this might end up being partially redundant. CC @drewstone @devpavan04

devpavan04:
@drewstone the error was due to an incorrect node version, which prevented me from serving the dapp locally for testing. Below is the connection info available to the dApp when connected to phantom. I have enabled phantom only for testing during this screen recording. Any other edge cases related to the Phantom wallet will be handled in the Solana routes PR when we actually use the solana chain.

https://github.com/user-attachments/assets/6775660d-cad3-4ca4-a0f9-209f98667a80

devpavan04:
- Remove Phantom wallet from the Wallet Connect modal when connecting from the dashboard page.

CleanShot 2025-04-07 at 06 18 08@2x

- Remove wallet explorer link icon if the wallet is incompatible with the active chain. This was a rare case encountered when testing the Phantom wallet connection without having the Solana chain. This is a temporary fix/check. I have removed the Phantom wallet option from the Wallet Connect modal, so this scenario will no longer occur.

CleanShot 2025-04-07 at 06 16 31@2x

- Always display the currently connected chain on all pages, instead of showing `Tangle Mainnet` or specific desired chains on respective pages. Previously, for example, if a user connected to `Base` on the Bridge page and then navigated back to the Dashboard page, they would see an alert when clicked prompting them to connect to the correct chain (`Tangle Mainnet`). Now, they see both the alert and the current chain they are connected to.

CleanShot 2025-04-07 at 06 18 58@2x

PR #2886, opened 2025-03-10
refactor(tangle-dapp): Use 0 instead of value placeholders

### LINKED ISSUE #2831: [TASK] Standardize No Data Value vs Zero Value Display
## Description 
We need to establish a standardized way to clearly make distinction between no data values and actual zero values across our UI.
### No Data (—)
- When the value is null/undefined
- When the data hasn't loaded yet
- When the value is not applicable for this context
- When there's an error fetching the data

### Zero (0)
- When we have successfully fetched the data and it's explicitly zero
- When a calculation results in zero

## Current Issues 
In the Restake Vaults table we have inconsistent usage:
- [x] "Available" column: Uses mix of '0' and '—' for empty values; should be '0'
- [x] "Deposits" column: Uses '—' for empty values, should be '0'
- [x] "TVL | Capacity" column: Uses '—' for empty values, should be '0'

In the account component;
- [x] Rewards and Points: Uses '—' for empty values, should be '0'

### Screenshot

Image

yurixander:
## Summary of changes

- 🔧 Small change: Tables & rewards should use `0` if there's no deposits or data, instead of `--` which should be used for loading/unavailable states.

### Proposed area of change

- [x] `apps/tangle-dapp`
- [ ] `apps/tangle-cloud`
- [ ] `libs/tangle-shared-ui`
- [ ] `libs/ui-components`

### Associated issue(s)

- Closes #2831

### Screen Recording

Screenshot 2025-03-10 at 01 11 37

### DIFF
--- a/apps/tangle-dapp/src/components/account/RewardsAndPoints.tsx
+++ b/apps/tangle-dapp/src/components/account/RewardsAndPoints.tsx
@@ -79,7 +79,7 @@ const RewardsAndPoints = () => {
             className="text-mono-140 dark:text-mono-40"
             component="span"
           >
-            {totalRewardsFormatted ?? EMPTY_VALUE_PLACEHOLDER}{' '}
+            {totalRewardsFormatted ?? 0}{' '}
           </Typography>
 
           <Typography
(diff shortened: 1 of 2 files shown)


AtelyPham (approved):
LGTM 👍

PR #2910, opened 2025-03-18
feat(tangle-cloud): implement logic of pending service request table

danielbui12:
## Summary of changes

- create `usePendingServiceRequest` hook
- update UI of Pending Instance table

### Proposed area of change

- [ ] `apps/tangle-dapp`
- [x] `apps/tangle-cloud`
- [x] `libs/tangle-shared-ui`
- [ ] `libs/ui-components`

### Associated issue(s)

- WIP https://github.com/tangle-network/dapp/issues/2556

### Screen Recording

https://github.com/user-attachments/assets/2fb23341-3e51-4cd2-9c16-04de7f10db89


libs/tangle-shared-ui/src/data/blueprints/usePendingServiceRequest.ts
```diff
@@ -0,0 +1,92 @@
...
+                const blueprintsData = blueprints
+                  .map((blueprint) => blueprint.unwrap())
+                  .map(([_, blueprint]) => toPrimitiveBlueprint(blueprint));
+                return createPendingServiceRequests(
+                  primitiveServiceRequests,
+                  blueprintsData,
+                );
+              }),
+            );
+          }),
```
yurixander, line 74: Going forward it may be worth considering avoiding rx ops like these because they tend to get complex and deeply nested quickly. Instead, I've found that breaking it down step by step using subsequent useMemos keeps things simpler. There should be some examples in other hooks about this approach. Rxjs map() should be fine though, but perhaps not other things like combineLatest, switchMap, etc.

yurixander: This specific case doesn't seem complex or deeply nested though.

libs/tangle-shared-ui/src/data/blueprints/utils/toPrimitiveService.ts
```diff
@@ -28,6 +34,77 @@ export default function toPrimitiveService({
...
+          approvalState.asApproved.securityCommitments.map((commitment) => {
+            return {
+              asset: createRestakeAssetId(commitment.asset),
+              exposurePercent: commitment.exposurePercent.toNumber(),
+            };
+          });
+        break;
+      default:
+        throw new Error('Unknown approval state type');
+    }
```
yurixander, line 102: Have you covered all cases? If so it'd be better to do something like this:


```
default:
      const _exhaustiveCheck: never = approvalState.type; // ❌ TypeScript compile-time error if a case is missing
      throw new Error(`Unknown approval state type: ${approvalState.type}`);
```

yurixander (changes requested):

libs/tangle-shared-ui/src/data/blueprints/utils/toPrimitiveService.ts
danielbui12: @yurixander how did you get that error?

yurixander: Not getting an error, rather I'm suggesting that approach because it throws a compile-time type error which is helpful because it lets the dev know that they're missing handling an extra case in the event where the node is updated and new cases are added. Without that exhaustive check, you'd only learn about the bug at runtime if you encounter the error.

danielbui12: I think I convered every cases, cuz following `tangle-substrate-type`, `approvalState` has only 3 types
```
        readonly type: 'Pending' | 'Approved' | 'Rejected';
```

danielbui12: If I added `const _exhaustiveCheck: never = approvalState.type;` , I would got error from eslint `'_exhaustiveCheck' is assigned a value but never used.eslint[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)`

yurixander: Try this: throw new Error(`Unhandled case: ${_exhaustiveCheck}`);. No hassle if it won't work though, its ok as it is

danielbui12: @yurixander  Tried `yarn build` and all good!

apps/tangle-cloud/src/pages/instances/Instances/PendingInstanceTable.tsx
```diff
@@ -201,39 +165,61 @@ export const PendingInstanceTable: FC<PendingInstanceTabProps> = ({
...
+                        ) ?? ''
+                      }
+                      target="_blank"
+                      rel="noopener noreferrer"
+                    >
+                      <Button variant="link" className="uppercase body4">
+                        {operatorIdentityMap?.get(owner)?.name ??
+                          shortenString(owner)}
+                      </Button>
+                    </Link>
```
AtelyPham, line 201: To render a link directly, you can pass `href` and `target` into the button component.

```tsx
<Button href={accountUrl} target='_blank' variant="link">

</Button>
```

libs/tangle-shared-ui/src/data/blueprints/usePendingServiceRequest.ts
```diff
@@ -0,0 +1,88 @@
...
+      );
+  }, [serviceRequestEntries, operatorAccountAddress]);
+
+  const blueprintIds = useMemo(() => {
+    return primitiveServiceRequests.map(
+      (serviceRequest) => serviceRequest.blueprint,
+    );
+  }, [primitiveServiceRequests]);
+
+  const { result: blueprints } = useApiRx(
```
AtelyPham, line 58: Combine the `loading` and `error` states with `...rest` and return the result as metadata for this hook.

AtelyPham (approved):
Left some comments, otherwise LG 👍

PR #2920, opened 2025-03-24
refactor(repo): move useAgnosticTx to common

danielbui12:
## Summary of changes

- refactor `useAgnosticTx` and relevant hooks to common 

### Proposed area of change

- [x] `apps/tangle-dapp`
- [ ] `apps/tangle-cloud`
- [ ] `apps/leaderboard`
- [x] `libs/tangle-shared-ui`
- [ ] `libs/ui-components`

### Associated issue(s)

- Closes

### Screen Recording

### DIFF
--- a/apps/tangle-dapp/src/data/liquidStaking/tangle/useLsPoolJoinTx.ts
+++ b/apps/tangle-dapp/src/data/liquidStaking/tangle/useLsPoolJoinTx.ts
@@ -3,10 +3,11 @@ import { useCallback } from 'react';
 
 import { TxName } from '../../../constants';
 import { PrecompileAddress } from '@tangle-network/tangle-shared-ui/constants/evmPrecompiles';
-import useAgnosticTx from '../../../hooks/useAgnosticTx';
-import { EvmTxFactory } from '../../../hooks/useEvmPrecompileCall';
-import { SubstrateTxFactory } from '../../../hooks/useSubstrateTx';
+import useAgnosticTx from '@tangle-network/tangle-shared-ui/hooks/useAgnosticTx';
+import { EvmTxFactory } from '@tangle-network/tangle-shared-ui/hooks/useEvmPrecompileCall';
+import { SubstrateTxFactory } from '@tangle-network/tangle-shared-ui/hooks/useSubstrateTx';
 import LST_PRECOMPILE_ABI from '@tangle-network/tangle-shared-ui/abi/lst';
+import { SUCCESS_MESSAGES } from '../../../hooks/useTxNotification';
 
 type Context = {
   poolId: number;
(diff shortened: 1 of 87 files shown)


danielbui12:
Initially, I only intended to refactor `useAgnosticTx` to make it more generic, but I found that we are using hooks in hook, which led to lots of changes.

apps/tangle-dapp/src/data/liquidStaking/tangle/useLsPoolJoinTx.ts
```diff
@@ -35,6 +36,7 @@ const useLsPoolJoinTx = () => {
     precompileAddress: PrecompileAddress.LST,
     evmTxFactory,
     substrateTxFactory,
+    successMessageByTxName: SUCCESS_MESSAGES,
```
yurixander, line 39: Smart move

apps/tangle-dapp/src/data/restake/useNativeRestakeUnstakeCancelTx.ts
```diff
@@ -55,6 +55,7 @@ const useNativeRestakeUnstakeCancelTx = () => {
     precompileAddress: PrecompileAddress.BATCH,
     evmTxFactory,
     substrateTxFactory,
+    successMessageByTxName: SUCCESS_MESSAGES,
```
yurixander, line 58: Could you make this the default in the fn params of useAgnosticTx? To avoid repeating it everywhere

yurixander (changes requested):
Great job! This was much needed. Now the other dapps can reuse these hooks. One small change and its good to go from me.

apps/tangle-dapp/src/data/restake/useNativeRestakeUnstakeCancelTx.ts
danielbui12: Cuz dapp and cloud are using different messages.
We can improve this by creating wrapped version in dapp and cloud separatedly

drewstone: I think he's just proposing moving it the `props` @danielbui12, why would we need a wrapped version? Move to props/params but set as default?

drewstone (approved):

apps/tangle-dapp/src/data/restake/useNativeRestakeUnstakeCancelTx.ts
danielbui12: @drewstone I understand his point, but it seems that dapp and cloud are utilizing different message values.

The `successMessageByTxName` is defined within `useTxNotification`, and it is used inside `useAgnosticTx`, which, in turn, is used within `useNativeRestakeUnstakeCancelTx`.

Assigning a default value to this property could potentially create confusion, I think

PR #2966, opened 2025-04-12
feat(tangle-dapp): Add Protocol Stats Component

AtelyPham:
## Summary of changes

**__Leaderboard:__**
- Implement animations and transitions.
- Add sliding number indexing indicator.

**__Tangle dApp:__**
- Integrate protocol stats component with real asset price data.
- Implement vault highlight slider component.
- Enhance vault table component:
    - Improve loading time.
    - Add assets column.
    - Implement EVM data caching.
    - Add vault ID (default sorting).
    - Refactor expandable asset table for scrollability with sticky headers.
    - Separate table and pie chart with cards for improved UI.
- Enhance network selection button:
    - Add loading indicator during network selection.
    - Improve display of selected network in dropdown.

**__Tangle Shared UI:__**
- Refactor API hooks for consistency and reusability.
- Refactor vault data into reusable hook.
- Implement reusable, cached token price fetching functions and hooks.

**__UI Components:__**
- Add `useMemorizedValue` hook for render optimization and to prevent infinite re-renders.
- Handle fractionMaxLength when using `formatDisplayAmount` with the `si` option.

### Proposed area of change

- [x] `apps/tangle-dapp`
- [x] `apps/tangle-cloud`
- [x] `apps/leaderboard`
- [x] `libs/tangle-shared-ui`
- [x] `libs/ui-components`

### Associated issue(s)

- Closes https://github.com/tangle-network/dapp/issues/2926
- Closes https://github.com/tangle-network/dapp/issues/2673

### Screen Recording

https://github.com/user-attachments/assets/2fb28fe0-fd46-4d01-93ae-7b9c3b8aafa5


apps/tangle-dapp/public/static/assets/vaults/vault-btc.png
yuri-xyz, line 1: We should optimize the size of these assets using something like https://tinypng.com/ to improve loading times, especially for users with slower connections.

libs/tangle-shared-ui/src/data/restake/useRestakeVaults.ts
```diff
@@ -0,0 +1,200 @@
...
+    },
+  ],
+]);
```
yuri-xyz, line 200: Structuring the icon like this in raw data: format feels a bit strange. Any better ideas for this? Can't it be as an SVG on its own file?

libs/tangle-shared-ui/src/hooks/useLocalStorage.ts
```diff
@@ -50,6 +51,11 @@ export type BridgeDestTxIds = Record<
   }
 >; // accountAddress -> { hyperlane: [] }
 
+export type TvlHistory = {
+  timestamp: number;
+  tvlInUsd: string;
+};
+
```
yuri-xyz, line 58: We should probably avoid storing stale data going forward, unless it takes a while to actually load it and it makes sense... for TVL do you think it's worth it?

yuri-xyz (approved):

libs/tangle-shared-ui/src/data/restake/useRestakeVaults.ts
AtelyPham: This structure mirrors the on-chain data structure, but it will be removed in the next chain upgrade; we will then fetch this data directly on-chain.

libs/tangle-shared-ui/src/hooks/useLocalStorage.ts
AtelyPham: To mitigate rate-limiting issues with the public API for token price fetching, I think we can store and display the TVL. To ensure transparency, we should include a timestamp indicating when the TVL was last calculated.

apps/tangle-dapp/public/static/assets/vaults/vault-btc.png
AtelyPham: Done ✅ Thanks for the suggestion!

PR #3094, opened 2026-01-14
fix(tangle-dapp): fix claim wallet connection and add unlock configuration display

vutuanlinh2k2:
## Summary
- Add substrate wallet selection modal for claim migration with proper wallet connection handling
- Add unlock configuration display to claim migration page
- Replace empty function with Function.prototype for lint compliance
- Code formatting fixes

## Images
CleanShot 2026-01-14 at 16 08 30@2x
CleanShot 2026-01-14 at 15 34 53@2x


tangletools:
Likely needs some small more improvements indicating that that is a tx hash?

PR #3095, opened 2026-01-26
feat: add operator delegation mode support and deprecate rewards hooks

vutuanlinh2k2:
## Summary
- Add operator delegation mode checks for restaking operations
- Display operator delegation mode in tables (both tangle-dapp and tangle-shared-ui)
- Deprecate rewards hooks pending RewardsManager integration
- Update ABI definitions for multiAssetDelegation and related contracts
- Remove TNT_CORE_CONTRACT_UPDATES tracking doc

## Video

https://github.com/user-attachments/assets/301e09e6-c220-4b3f-9496-a8195e12caaf

_(note: the flicker in the operator table at the end of the video is already fixed)_

## Test plan
- [ ] Verify operator delegation mode is displayed correctly in operator tables
- [ ] Test delegate page with operators in different delegation modes
- [ ] Confirm deprecated rewards hooks are properly marked and don't break existing functionality

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vutuanlinh2k2:
## Response to Claude's Review

Thanks for the detailed review! I've analyzed each issue raised and here's my assessment:

### Issues We'll Address

1. **Stale time inconsistency** ✅ - Valid. `useOperatorDelegationModes` uses 60s while other hooks use 30s. Will standardize.

2. **Map key casing** ✅ - Valid edge case. Will add `.toLowerCase()` normalization when looking up delegation info to prevent potential mismatches.

3. **BigInt parsing safety** ✅ - Valid. Will add try-catch wrapper for `BigInt()` conversions in GraphQL parsing to handle unexpected data gracefully.

### Issues That Are Actually Fine

4. **Type safety in DelegationMode handling** - This is intentional defensive coding. The enum only has 3 values (0=Disabled, 1=Whitelist, 2=Open), and defaulting unknown values to `Disabled` is the safest behavior as it prevents unauthorized delegation.

5. **"Race condition" in delegation loading** - This is a documented design choice (see comment on line 255: "canDelegate undefined means we don't have the info yet (backwards compat)"). Buttons remain enabled while loading for backwards compatibility.

6. **Error boundary missing** - The hooks properly handle errors via `Promise.allSettled` and expose errors through React Query's `error` property. Error boundaries are a component-level concern.

7. **Incomplete deprecation (ERROR status)** - This is intentional. The hook is deprecated and *should* signal failure. The JSDoc clearly documents this behavior.

8. **Type casting (`as Promise<...>`)** - Standard pattern when using viem's `readContract` without generated typed contracts. Works correctly.

### Low Priority (Won't Fix Now)

9. **Excessive contract calls** - Valid observation but standard practice. Will optimize with multicall if performance becomes an issue.

10. **Magic numbers** - Minor style preference, low impact.

Will push fixes for items 1-3 shortly.

vutuanlinh2k2:
@claude please review my recent changes

PR #3132, opened 2026-02-19
feat(tangle-cloud): runtime permitted callers management

vutuanlinh2k2:
## Summary
- Add permitted callers management UI to the service detail page, allowing service owners to add/remove addresses that can submit jobs at runtime
- Fetch `permittedCallers` from the GraphQL indexer with backward-compatible fallback for older indexer schemas
- Rename "View" button to "Blueprint" in running instance table for clarity

## Video
https://github.com/user-attachments/assets/38264975-c6e5-418a-a800-053e1b060987

## Test plan
- [ ] Verify service detail page loads permitted callers list correctly
- [ ] Verify owner can add a new permitted caller address
- [ ] Verify owner can remove a permitted caller
- [ ] Verify non-owners see a read-only view with an info message
- [ ] Verify input validation (zero address, duplicate, owner address, invalid format)
- [ ] Verify backward compatibility when indexer doesn't expose `permittedCallers` field

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/tangle-cloud/src/pages/instances/Instances/RunningInstanceTable.tsx
+++ b/apps/tangle-cloud/src/pages/instances/Instances/RunningInstanceTable.tsx
@@ -234,7 +234,7 @@ export const RunningInstanceTable: FC<RunningInstanceTableProps> = () => {
                     variant="utility"
                     className="uppercase body4 bg-blue-10 dark:bg-blue-120 text-blue-70 dark:text-blue-40 hover:bg-blue-20 dark:hover:bg-blue-110 border border-blue-30 dark:border-blue-100 transition-all duration-200"
                   >
-                    View
+                    Blueprint
                   </Button>
                 </Link>
 
--- a/apps/tangle-cloud/src/pages/instances/Instances/RunningInstanceTable.tsx
+++ b/apps/tangle-cloud/src/pages/instances/Instances/RunningInstanceTable.tsx
@@ -1,11 +1,4 @@
-import {
-  useMemo,
-  useState,
-  useCallback,
-  type FC,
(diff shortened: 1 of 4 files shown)


vutuanlinh2k2:
### Addressing Claude's Review Comments

The latest commit (`288a578`) addresses the actionable items from the review:

**Fixed:**
- **Unused props in `RunningInstanceTable`** — Removed `RunningInstanceTableProps` interface and the unused `refreshTrigger`/`setRefreshTrigger` props from the component and its call site.
- **Loading state for permitted callers** — Added `SkeletonLoader` to the permitted callers card while `isLoadingOnChainDetails` or `isLoadingServices` is true.
- **`aclHistoryTxChecks` removed entirely** — The tx history merging (`useQueries` + `useTxHistoryStore` + `usePublicClient`) was unnecessary since `onSuccess` callbacks in both `useAddPermittedCallerTx` and `useRemovePermittedCallerTx` already refetch indexer data. This also eliminates the `staleTime: 30_000` magic number and the N RPC calls per render that the review flagged.

**Review points that don't need action:**
- *"Memory usage — permittedCallers recreates arrays on every render"* — Incorrect; the computation is already wrapped in `useMemo`.
- *"Race conditions from concurrent refetchPermission/refetchService"* — These are independent queries with no shared state; parallel refetch is correct.
- *"Debouncing input validation"* — The validation is purely local (in-memory string comparisons via viem), no network calls involved.
- *"Add JSDoc comments"* — Project guidelines say to avoid comments unless logic is complex.
- *"staleTime: 30_000 in useServices.ts"* — Outside the scope of this PR.

PR #3148, opened 2026-03-21
fix(tangle-cloud): resolve round-2 payment review findings

drewstone:
## Summary

Fixes all findings from the codex/claude ensemble review on #3147.

### Write queue fix
- notesRef.current updated synchronously inside the promise chain so back-to-back addNote calls see each other's results (was dropping notes)

### Credit key lifecycle
- saveCreditKeys now requires encryption key (throws if missing — no more plaintext storage)
- StoredCreditKeys gains isLocked field — locked accounts clearly marked in UI
- deleteCreditKeys verifies owner address before deleting (was cross-wallet deletable)
- FundCreditsContainer requires unlocked keypair, removed misleading proof progress
- SpendAuthContainer rejects signing with locked/empty private keys

### Wallet switch safety
- CreditsProvider uses generation counter for stale-response guard
- useKeypair resets isLoading on address change (prevents permanently stuck state)

### IDB
- onversionchange handler closes and invalidates stale connections across tabs

### Layout regression fix
- PageLayout defaults to md:px-5 (original instances/rewards/earnings padding)
- blueprints/operators layouts explicitly pass md:px-8 lg:px-10 relative

### UX honesty
- Withdraw shows confirmed-notes balance only, not total (unconfirmed notes aren't spendable)
- AmountInput MAX works correctly for 0n balance

## Test plan
- [x] yarn nx typecheck tangle-cloud passes
- [x] yarn nx lint tangle-cloud passes (0 errors, 0 warnings)
- [x] yarn nx test tangle-cloud passes (52/52)

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -50,6 +50,10 @@ https://stats.tangle.tools
 # Something happened with conventional commits link, temporary disabled to fix the CI
 https://www.conventionalcommits.org/en/v1.0.0/
 
+# External sites that block bots or have intermittent Cloudflare errors
+https://www.tangle.tools/
+https://openai.com/index/harness-engineering/
+
 # Files
 /**/CHANGELOG.md
 
--- a/apps/tangle-cloud/src/app/CreditsProvider.tsx
+++ b/apps/tangle-cloud/src/app/CreditsProvider.tsx
@@ -7,6 +7,7 @@ import {
   useContext,
   useEffect,
   useMemo,
+  useRef,
   useState,
 } from 'react';
 import { useAccount } from 'wagmi';
@@ -25,6 +26,8 @@ interface CreditsContextValue {
   generateAndStoreCreditKeys: (label?: string) => Promise<StoredCreditKeys>;
   removeCreditAccount: (commitment: string) => Promise<void>;
   isLoading: boolean;
+  /** True when keypair is unlocked and credit keys are decryptable */
+  isUnlocked: boolean;
 }
 
 const CreditsContext = createContext<CreditsContextValue | null>(null);
@@ -35,34 +38,39 @@ export const CreditsProvider: FC<PropsWithChildren> = ({ children }) => {
   const [creditAccounts, setCreditAccounts] = useState<StoredCreditKeys[]>([]);
   const [isLoading, setIsLoading] = useState(true);
 
+  // Generation counter to detect stale async loads
+  const genRef = useRef(0);
+
   const encryptionKey = keypair
     ? keccak256(
         encodePacked(['string'], [keypair.privateKey + ':credit-encryption']),
       )
     : undefined;
 
-  // Reload when address or encryption key changes
   useEffect(() => {
     setCreditAccounts([]);
     setIsLoading(true);
+    const gen = ++genRef.current;
 
     if (!address) {
       setIsLoading(false);
       return;
     }
 
-    const currentAddress = address;
     const load = async () => {
       try {
-        const keys = await loadCreditKeysForAddress(
-          currentAddress,
-          encryptionKey,
-        );
-        setCreditAccounts(keys);
+        const keys = await loadCreditKeysForAddress(address, encryptionKey);
+        if (genRef.current === gen) {
+          setCreditAccounts(keys);
+        }
       } catch {
-        setCreditAccounts([]);
+        if (genRef.current === gen) {
+          setCreditAccounts([]);
+        }
       }
-      setIsLoading(false);
+      if (genRef.current === gen) {
+        setIsLoading(false);
+      }
     };
     load();
   }, [address, encryptionKey]);
@@ -70,6 +78,11 @@ export const CreditsProvider: FC<PropsWithChildren> = ({ children }) => {
   const generateAndStoreCreditKeys = useCallback(
     async (label?: string): Promise<StoredCreditKeys> => {
       if (!address) throw new Error('Wallet not connected');
+      if (!encryptionKey) {
+        throw new Error(
+          'Unlock your shielded keypair before generating credit keys',
+        );
+      }
 
       const privateKey = generatePrivateKey();
       const account = privateKeyToAccount(privateKey);
@@ -89,6 +102,7 @@ export const CreditsProvider: FC<PropsWithChildren> = ({ children }) => {
         salt,
         label,
         createdAt: Date.now(),
+        isLocked: false,
       };
 
       await saveCreditKeys(keys, address, encryptionKey);
@@ -98,12 +112,16 @@ export const CreditsProvider: FC<PropsWithChildren> = ({ children }) => {
     [address, encryptionKey],
   );
 
-  const removeCreditAccount = useCallback(async (commitment: string) => {
-    await deleteCreditKeys(commitment);
-    setCreditAccounts((prev) =>
-      prev.filter((k) => k.commitment !== commitment),
-    );
-  }, []);
+  const removeCreditAccount = useCallback(
+    async (commitment: string) => {
+      if (!address) return;
+      await deleteCreditKeys(commitment, address);
+      setCreditAccounts((prev) =>
+        prev.filter((k) => k.commitment !== commitment),
+      );
+    },
+    [address],
+  );
 
   const value = useMemo<CreditsContextValue>(
     () => ({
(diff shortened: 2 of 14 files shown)


drewstone:
## ❌ PR Review w/ codex, claude

> Don't merge. High: IndexedDB save failure poisons write queue and diverges in-memory state from persisted state in `apps/tangle-cloud/src/app/ShieldedProvider.tsx:99`. Medium: In-flight credit-key generation can append into the next wallet's state in `apps/tangle-cloud/src/app/CreditsProvider.tsx:106`. Medium: generateAndStoreCreditKeys can pollute state after address switch in `apps/tangle-cloud/src/app/CreditsProvider.tsx:109`. Plus 6 more findings.

| | |
|---|---|
| **Recommendation** | Needs Work (39/100) |
| **Findings** | 9 total — 🔴 1 high, 🟠 5 medium, 🟡 3 low |
| **Ensemble** | 2 reviewers × 3 tracks |
| **Files reviewed** | 14 files changed |
| **Validator** | claude |
| **Provenance** | [individual reviewer outputs](https://gist.github.com/drewstone/f01a5c20f6377b451530795b32bafc6d) |

---

### 🔴 HIGH (1)

- **IndexedDB save failure poisons write queue and diverges in-memory state from persisted state** [`apps/tangle-cloud/src/app/ShieldedProvider.tsx:99`](https://github.com/tangle-network/dapp/blob/c7df1535917c12d9e384fb76fdd2422c2bb91d7a/apps/tangle-cloud/src/app/ShieldedProvider.tsx#L99)
  > The `persist` callback chains writes via `writeQueueRef.current.then(async () => { ... })`. If `storageRef.current.save()` rejects, the returned promise rejects, which means `writeQueueRef.current` holds a rejected promise. Every subsequent `.then()` chained on it will be skipped — the entire write queue is dead. No future `addNote`, `removeNote`, or `importNotes` call will execute.

Before this PR, `setNotes(updated)` happened AFTER the save, so a save failure at least kept in-memory state cons…

### 🟠 MEDIUM (5)

- **In-flight credit-key generation can append into the next wallet's state** [`apps/tangle-cloud/src/app/CreditsProvider.tsx:106`](https://github.com/tangle-network/dapp/blob/c7df1535917c12d9e384fb76fdd2422c2bb91d7a/apps/tangle-cloud/src/app/CreditsProvider.tsx#L106)
  > `CreditsProvider` protects the reload effect with `genRef`, but `generateAndStoreCreditKeys` still captures `address`/`encryptionKey`, awaits `saveCreditKeys(...)`, and then unconditionally does `setCreditAccounts((prev) => [...prev, keys])`. If the user starts generation on wallet A and switches wallets or changes lock state before the async save finishes, the old request can repopulate the current provider state with A's newly generated account. The stored record is written under the old owner…

- **generateAndStoreCreditKeys can pollute state after address switch** [`apps/tangle-cloud/src/app/CreditsProvider.tsx:109`](https://github.com/tangle-network/dapp/blob/c7df1535917c12d9e384fb76fdd2422c2bb91d7a/apps/tangle-cloud/src/app/CreditsProvider.tsx#L109)
  > The genRef guard correctly prevents stale *loads* from overwriting state, but generateAndStoreCreditKeys has no equivalent guard. If the user switches wallets while key generation is in-flight, the sequence is: (1) generateAndStoreCreditKeys captures address=A, encryptionKey=K1 via closure; (2) user switches to address B; (3) effect runs: setCreditAccounts([]), increments genRef, starts loading address B's keys; (4) the in-flight generate completes and calls setCreditAccounts(prev => [...prev, k…

- **Create-account warning uses stored-keypair state instead of actual unlocked state** [`apps/tangle-cloud/src/containers/payments/FundCreditsContainer.tsx:57`](https://github.com/tangle-network/dapp/blob/c7df1535917c12d9e384fb76fdd2422c2bb91d7a/apps/tangle-cloud/src/containers/payments/FundCreditsContainer.tsx#L57)
  > `FundCreditsContainer` shows its unlock warning only when `!hasDerivedKeypair`, but it disables the label input and submit button when `!isUnlocked`. Those are no longer the same condition: `CreditsProvider.isUnlocked` is only true when there is a live `encryptionKey` from an unlocked shielded keypair, while `useKeypair` still marks the address as having a keypair as soon as it finds a saved entry in local storage. In the common case where a user has previously derived a shielded keypair but has…

- **SpendAuthContainer lets users select locked accounts and fill out the entire form before discovering the account is locked** [`apps/tangle-cloud/src/containers/payments/SpendAuthContainer.tsx:224`](https://github.com/tangle-network/dapp/blob/c7df1535917c12d9e384fb76fdd2422c2bb91d7a/apps/tangle-cloud/src/containers/payments/SpendAuthContainer.tsx#L224)
  > The locked-account guard at line 77 only fires inside handleSign — after the user has selected a locked account, filled in service ID, job index, amount, operator, and expiry. There's no visual indication in the dropdown or form that the selected account is locked. The user discovers this only after clicking 'Sign Authorization'. This is a valid-user trap: someone who reconnects their wallet (keypair not yet unlocked) sees their old credit accounts listed as if they're usable, fills out the whol…

- **Failed note save now leaves optimistic state applied and poisons the write queue** [`apps/tangle-cloud/src/app/ShieldedProvider.tsx:101`](https://github.com/tangle-network/dapp/blob/c7df1535917c12d9e384fb76fdd2422c2bb91d7a/apps/tangle-cloud/src/app/ShieldedProvider.tsx#L101)
  > `updateQueuedNotes` now mutates `notesRef` and React state before awaiting IndexedDB persistence, but the queue still has no rejection recovery. If `storageRef.current.save(...)` rejects once, the user sees the new notes in memory even though they were never durably stored, and `writeQueueRef.current` becomes a rejected promise. Because later calls chain with `.then(...)` and no rejection handler, every subsequent queued note update is skipped. In practice, one storage error can make shielded no…

### 🟡 LOW (3)

- **onversionchange closes DB under in-flight callers without recovery** [`apps/tangle-cloud/src/utils/payments/db.ts:44`](https://github.com/tangle-network/dapp/blob/c7df1535917c12d9e384fb76fdd2422c2bb91d7a/apps/tangle-cloud/src/utils/payments/db.ts#L44)
  > When another tab triggers a schema upgrade, the onversionchange handler calls db.close() and nullifies dbPromise. Any code that already obtained the db reference from a prior openDb() call and has an in-flight or upcoming transaction will get an InvalidStateError. The next openDb() call will reconnect, but there's no retry or error propagation to the caller that was mid-operation. In practice this is rare (only happens during version bumps with multiple tabs), but a failed mid-operation save or …

- **FundCreditsContainer warning alert uses hasDerivedKeypair but button uses isUnlocked — different source, same semantics today, fragile coupling** [`apps/tangle-cloud/src/containers/payments/FundCreditsContainer.tsx:57`](https://github.com/tangle-network/dapp/blob/c7df1535917c12d9e384fb76fdd2422c2bb91d7a/apps/tangle-cloud/src/containers/payments/FundCreditsContainer.tsx#L57)
  > Line 57 shows the warning alert when `!hasDerivedKeypair` (from ShieldedProvider, meaning keypair === null). Line 91 disables the button when `!isUnlocked` (from CreditsProvider, meaning encryptionKey is falsy, which is derived from keypair being null). Today these are equivalent because encryptionKey is derived directly from keypair.privateKey. But the component pulls from two different providers for what is conceptually the same gate. If either derivation changes independently (e.g., CreditsPr…

- **SpendAuthContainer button disabled check doesn't account for locked accounts** [`apps/tangle-cloud/src/containers/payments/SpendAuthContainer.tsx:339`](https://github.com/tangle-network/dapp/blob/c7df1535917c12d9e384fb76fdd2422c2bb91d7a/apps/tangle-cloud/src/containers/payments/SpendAuthContainer.tsx#L339)
  > The Sign Authorization button's disabled condition (line 339-346) checks for selectedAccount truthiness but not selectedAccount.isLocked. A user can select a locked account and the button remains enabled. The runtime guard at line 77 catches this and shows an error, so it's not a functional bug, but the button should be disabled for locked accounts to match the pattern used in FundCreditsContainer where the button is proactively disabled rather than relying on runtime error messages.


---

### 🎯 What would get this approved

- [ ] Fix [high] IndexedDB save failure poisons write queue and diverges in-memory state from persisted state in apps/tangle-cloud/src/app/ShieldedProvider.tsx

---
<sub>pr-reviewer v0.5.0 · review #1 · 2026-03-21T01:04:11.842126+00:00</sub>

PR #3219, opened 2026-05-21
fix(blueprint-metadata): resolve bare github.com URIs to raw metadata JSON

tangletools:
## Summary

- Several blueprints just registered on Base Sepolia (e.g. ai-agent-sandbox, llm-inference, modal-inference, image-gen-inference) carry an on-chain `metadataUri` that points at the repo's GitHub HTML page (`https://github.com/<owner>/<repo>`). The dapp `fetch()`-es that URL expecting JSON, receives HTML, `parseBlueprintMetadataJsonText` bails out, and every affected blueprint renders as unavailable.
- Teach `resolveBlueprintMetadataFetchUrl` to translate a bare `github.com/<owner>/<repo>` URL (optionally with a trailing slash or `.git` suffix) into `raw.githubusercontent.com/<owner>/<repo>/main/metadata/blueprint-metadata.json`. `ipfs://` handling is untouched, URLs with sub-paths (`/tree/main/...`) are passed through unchanged, and any other HTTPS URL still flows through `rewriteLocalhostUrlForBrowser`.
- Repos that default to `master` instead of `main` will 404 for now — that's accepted as a follow-up (fixable on-chain via `updateBlueprint`, or we can add a `main` → `master` fallback later). The four repos with on-chain registrations today all default to `main`.
- The resolver moves into its own `metadataFetchUrl.ts` so unit tests don't transitively pull in the authoring module's `import.meta.env` reference; `authoring.ts` re-exports the function so the public API is identical.

## Test plan

- [x] `yarn nx test tangle-shared-ui --testPathPattern=authoring` — 6 new cases (ipfs, bare GitHub, trailing slash, `.git`, sub-path passthrough, unrelated HTTPS passthrough), all pass.
- [x] `yarn nx test tangle-shared-ui` — full suite green (32 tests).
- [x] `yarn nx test tangle-cloud` — full suite green (109 tests).
- [x] `yarn nx lint tangle-shared-ui` — clean.
- [x] `yarn nx build tangle-cloud` — clean.
- [ ] Manual smoke on Base Sepolia: hit the four affected blueprint detail pages and confirm metadata loads.

PR #3222, opened 2026-05-21
fix(tangle-cloud): chain-read fallback when Envio indexer is absent

tangletools:
## The bug

The blueprints listing in tangle-cloud was **empty on Base Sepolia even though 13 blueprints are registered on-chain** — because both `useBlueprints` and `useBlueprintsWithMetadata` are GraphQL-only via the Envio indexer, and there is currently no hosted Envio indexer for Base Sepolia (`VITE_ENVIO_TESTNET_ENDPOINT` is unset on staging).

User-visible symptom: navigated to /blueprints, saw "0 Catalog", didn't know any of the 13 registered blueprints existed.

## The fix

Add a `fetchBlueprintsOnChain` helper that reads the registry directly from the Tangle contract: `blueprintCount()` followed by a batched `getBlueprint(id)` + `getBlueprintDefinition(id)` per entry. Wire both hooks to use it as a fallback when:

1. the indexer call throws (endpoint missing / unreachable), OR
2. the indexer returns 0 rows while the chain has live blueprints

The fallback is bounded by the existing `limit` arg (default 100). At the current scale of 13 entries it's tiny; safe up to the next couple orders of magnitude on testnet. Production-scale (1000s of blueprints) still wants a real indexer.

The chain-read path populates the same `Blueprint` shape the GraphQL path returns, so downstream consumers don't branch on data source. Fields the indexer tracks but the chain doesn't expose (`updatedAt`) get a sensible default (`= createdAt`) so sort order stays predictable.

## After this lands

`develop.cloud.tangle.tools` rebuilds, the blueprints listing fills with the 13 registered entries, and (with the GitHub-URL metadata resolver from #3219/#3220 already shipped) each one renders with its metadata from the repo's `metadata/blueprint-metadata.json`.

Once Envio is stood up against the deployed Base Sepolia contracts and `VITE_ENVIO_TESTNET_ENDPOINT` is set in Netlify env, the GraphQL path becomes primary again automatically and the chain-read fallback quietly disengages.

## Test plan

- [x] `yarn nx build tangle-cloud --skip-nx-cache` clean
- [ ] Manual: load `develop.cloud.tangle.tools/blueprints` after merge → see all 13 entries
- [ ] Manual: load `develop.cloud.tangle.tools/instances` → RegisteredBlueprints section populated
- [ ] Once a hosted Envio is configured, verify GraphQL becomes primary again

### DIFF
--- a/apps/tangle-cloud/src/app/app.tsx
+++ b/apps/tangle-cloud/src/app/app.tsx
@@ -1,6 +1,7 @@
 import { Navigate, Route, Routes } from 'react-router';
 import { lazy, Suspense, type FC, type ReactNode } from 'react';
 import Layout from '../components/Layout';
+import ScrollToTop from '../components/ScrollToTop';
 import Providers from './providers';
 import { PagePath } from '../types';
 import { Skeleton } from '@tangle-network/sandbox-ui/primitives';
@@ -70,6 +71,7 @@ const withLayout = (LayoutCmp: FC<{ children: ReactNode }>, Page: FC) => (
 const App: FC = () => {
   return (
     <Providers>
+      <ScrollToTop />
       <Layout>
         <Routes>
           <Route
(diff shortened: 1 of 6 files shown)


drewstone (approved):
Approved.

PR #3242, opened 2026-05-25
fix(tangle-cloud): sync wallet to iframe on handshake + iframe fills its host

drewstone:
Two embed bugs surfaced by the LLM Inference blueprint app.

## 1. Iframe stuck at "no-wallet" (no handoff)
The parent broadcasts `accountChanged` / `chainChanged` / `serviceContext` only from **mount effects**, which fire before the iframe has loaded and attached its `message` listener. The handshake handler acked but **didn't replay state**, so a late-handshaking iframe received the ack (and the SDK's client stops its handshake retry on ack) yet never got the account → permanent `no-wallet`.

Now the handshake handler replays the current wallet + chain + service context from a live `syncRef` right after acking. This is the parent-side mirror of the client's handshake-retry hardening.

## 2. Iframe left dead space / page scrolled
The frame had a hard `minHeight: 720px` and no `height: 100%`, so on a tall viewport it sat short inside its `calc(100vh-108px)` container (gap below), and the page's `pb-10` added more — making the **parent** scroll.

- Frame now fills its host: `height: 100%`, rounded (`12px`), no min-height floor.
- Host container gets a small padding gutter (`p-2 md:p-3`) + `overflow-hidden`, so the rounded iframe reads as a contained surface filling most of the parent.
- Layout reclaims the page's bottom padding (`-mb-10`) so the parent doesn't scroll — only the app inside the iframe scrolls.

Shared by all iframe blueprint apps (AI Trading, AI Agent Sandbox, LLM Inference).

## Test plan
- [x] `tsc --noEmit -p tsconfig.app.json` clean; full pre-push (lint/test/build) passed
- [ ] Visual: embed an iframe blueprint in Tangle Cloud — wallet shows connected; frame fills the area with a small rounded gutter; no parent scroll

### DIFF
--- a/apps/tangle-cloud/src/blueprintApps/components/BlueprintAppFrame.tsx
+++ b/apps/tangle-cloud/src/blueprintApps/components/BlueprintAppFrame.tsx
@@ -102,11 +102,13 @@ const BlueprintAppFrameInner = (
       loading="lazy"
       className={className}
       style={{
-        // Default to a sensible aspect; consumer can override.
+        // Fill the host container exactly — the layout owns the height, the
+        // app inside the iframe owns its own scroll. (A hard minHeight here
+        // left dead space below the frame on tall viewports.)
         width: '100%',
-        minHeight: '720px',
+        height: '100%',
         border: '0',
-        borderRadius: '16px',
+        borderRadius: '12px',
         ...style,
       }}
       {...iframeProps}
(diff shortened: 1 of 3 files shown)


tangletools (approved):
Handshake-replay is the correct parent-side mirror of the client retry; iframe fill + no-parent-scroll matches the shared layout. typecheck + pre-push green. Approving.

PR #3244, opened 2026-05-25
fix(tangle-cloud): kill unwanted italics on form controls + chips

drewstone:
The existing de-italic rule (`styles.css`) only targeted `button/a/span/p/h1–h6` and had no `!important`, so slanted **fallback** glyphs leaked through on the **search bar** (`<input>`) and **category chips**. Broadens the selector to form controls (`input/select/textarea/label/li/div`) and forces `font-style: normal !important`. Italics stay opt-in via the explicit `.italic` class.

## Test plan
- [x] full pre-push (lint/test/build) passed
- [ ] Visual: search bar + category chips render upright (no slant)


tangletools (approved):
Broadened de-italic with !important; opt-in via .italic preserved. Approving.

PR #3278, opened 2026-06-22
[RELEASE] feat(tangle-cloud): adopt parent-bridge wallet model for bazaar

drewstone:
## Summary

Inference Bazaar currently runs its **own** ConnectKit wallet inside the iframe (the only one of the three first-party iframe apps that doesn't use the parent bridge). This flips bazaar to the same parent-bridge model as trading-arena and agent-sandbox, so when embedded in Tangle Cloud the connected wallet flows from the parent (browser extensions can't inject into a sandboxed cross-origin iframe anyway — that's why the bridge exists).

### Change
Registry policy for `bazaar` (`apps/tangle-cloud/src/blueprintApps/registry.ts`):
- `allowReadAccount`: false → **true** (surfaces the parent wallet for read-only views)
- `allowChainSwitch`: false → **true** (restricted to Base Sepolia 84532, where the bazaar settlement contracts live)
- `allowPopups` + `allowSameOrigin`: **retained** (own-origin storage + explorer links; the wallet no longer depends on them once the bridge is active)
- `contracts` / `messages` remain `[]` — bazaar doesn't issue `signTransaction` yet

Companion test updated to lock in the new policy.

### Sequencing — important
This is the **prerequisite** for the bazaar-side connector swap (separate repo, `tangle-network/inference-bazaar`). Ship order matters:
1. **This PR first** (forward-compatible — shipping it alone changes nothing for users until bazaar deploys its connector swap, since bazaar won't exercise the bridge until then).
2. Then bazaar's `web3.tsx` swap (`detectTangleCloudParentOrigin` + `parentBridgeConnector`, mirroring arena/agent-sandbox) — **only after this reaches prod**, otherwise embedded bazaar would swap out ConnectKit while the parent still rejects `readAccount`, leaving it walletless.

## Gates
typecheck, lint (0 err), test (167 passed), build tangle-cloud all green.

### DIFF
--- a/apps/tangle-cloud/src/blueprintApps/registry.spec.ts
+++ b/apps/tangle-cloud/src/blueprintApps/registry.spec.ts
@@ -41,10 +41,13 @@ describe('blueprint app registry', () => {
     expect(trainingBySlug?.slug).toBe('training');
     expect(surplusByMetadata?.slug).toBe('bazaar');
     expect(surplusByMetadata?.manifest.externalApp?.mode).toBe('iframe');
-    // Surplus runs its own wallet in-frame, so it opts into allow-same-origin
-    // (cross-origin → its own storage) but not the parent bridge grants.
+    // Surplus now uses the parent-bridge wallet model (like trading-arena /
+    // agent-sandbox): allowReadAccount + allowChainSwitch granted so the
+    // connected wallet flows from the parent. allowSameOrigin is retained
+    // for own-origin app-state storage.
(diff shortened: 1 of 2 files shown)


tangletools (approved):
## ✅ Auto-approved PR — `30ac82f2`

Blanket team auto-approval is enabled for this reviewer service.
The full PR reviewer audit still runs separately and will publish findings if it detects issues.

<sub>tangletools · auto-approval · reason: `blanket_auto_approve` · 2026-06-22T08:07:25Z</sub>

